PR #13, opened 2025-07-08
Add bitmap milling

thomasmfish:
Add bitmap milling capabilities to fibsem-os. This is a continuation of https://github.com/DeMarcoLab/fibsem/pull/345 with updates for the more recent fibsem-os changes.

Key decisions:
- Bitmaps will be opened and converted to a consistent format by `FibsemBitmapSettings`, as this avoids requiring that all bits of code that might use it (e.g. plotting, milling) can handle files and arrays of both types. This also means that volume can more accurately be calculated.
- `BitmapPattern` and `TrenchBitmapPattern` will have both a `path` and an `array` attribute, which will be handled by `FibsemBitmapSettings`. `array` will currently override `path` if both are given.
- I've added a `_hidden_attributes` class variable to `BasePattern`, allowing array attributes to be hidden, since they can't be set by the GUI.
- `TrenchBitmapPattern` also has a `lower_path` and `lower_array` attribute to allow different pattterns on the bottom (defaults to using the same for both). This is necessary as bitmap resolution affects milling, so different trench heights could cause problems. An alternative method would be to join the trench height attributes but allowing for some customisation of top and bottom (e.g. fillet-like behaviour) could be useful in the future.

Required changes:
- Image/array based pattern plotting.


thomasmfish:
I've added a `draw_line_shape` function to complete `draw_pattern_shape`, but this does highlight the potential issue with using image-based plotting, as the line could easily be invisible for less zoomed in views. This led me to looking into patch-based plotting again.

I've tried using `PatchCollection`s again with what I hope is a fix for the legend (based on [this](https://matplotlib.org/stable/users/explain/axes/legend_guide.html#implementing-a-custom-legend-handler)). I haven't tested this properly yet, that's the next step.

thomasmfish:
Hi @patrickcleeve2, this isn't ready for review yet but I am working on the bitmap milling again, so thought I'd check in and see where things are at. Is there anything I should be aware of in terms of planned changes or otherwise?

Also, I'm happy to leave them until review but there are a couple of small commits that could be split from this PR if they would be useful to include sooner:
- bf36269
- ca3beaf

thomasmfish:
I have hit a problem in `draw_milling_patterns_in_napari`, as `add_shapes` doesn't appear to allow for images, and the image returned is both an ndarray and tuple of coordinates. I'm not sure what the original plan was around this, but I've updated how `draw_milling_patterns_in_napari` works so that it's a bit easier to handle images separately if necessary.

patrickcleeve2:
> I have hit a problem in `draw_milling_patterns_in_napari`, as `add_shapes` doesn't appear to allow for images, and the image returned is both an ndarray and tuple of coordinates. I'm not sure what the original plan was around this, but I've updated how `draw_milling_patterns_in_napari` works so that it's a bit easier to handle images separately if necessary.

Hi @thomasmfish, I don't have a nice solution for this at the moment. Currently, what I've done is explicitly define special cases that can't be drawn as shapes (Annulus, Bitmap), and create a separate image/label layer for these. It works ok, but it's pretty messy. Since the actual shapes are controlled by pattern dimension (meaning user can't drag shape around etc) it isn't an issue to not have shapes layers. 

Given the current limitations of the shapes layer (e.g. no annulus, no exclusion zone etc), and the need for bitmaps, i've done some work on drawing the patterns as image masks. The current utils should support drawing all the patterns as masks. For bitmaps, it would have to be drawn as an image layer, but since the labels/image layers are very similar it would reduce the complxity of the whole drawing func a lot. I think it will be much faster as well.

```python
import napari
from fibsem.ui.napari.patterns import COLOURS
from fibsem.milling.patterning.utils import create_pattern_mask
from autolamella.structures import AutoLamellaProtocol
from autolamella.config import PROTOCOL_PATH
import numpy as np 

protocol = AutoLamellaProtocol.load(PROTOCOL_PATH)

stages = protocol.milling["mill_rough"]
stages.extend(protocol.milling["mill_polishing"])
# stages.extend(protocol.milling["notch"])
stages.extend(protocol.milling["microexpansion"])
stages.extend(protocol.milling["fiducial"])

image = FibsemImage.generate_blank_image(hfw=stages[0].milling.hfw, resolution=(3072, 2048), random=True)
mask = np.zeros(image.data.shape, dtype=np.uint8)
colormap = {0: 'black'}
for i, stage in enumerate(stages):
    m = create_pattern_mask(stage, image)
    mask[m > 0] = i + 1
    colormap[i + 1] = COLOURS[i % len(COLOURS)]

viewer = napari.Viewer()
viewer.add_image(image.data, name="Image", colormap='gray', blending='additive')
viewer.add_labels(mask, name="Mask", colormap=colormap)
# viewer.add_image(mask*25, name="Mask Image", colormap='gray', blending='additive')

napari.run()

```

thomasmfish:
@patrickcleeve2 thanks for your reply, I'll try to sort those commits tomorrow.

I did briefly get the bitmaps _mostly_ working (see the screenshot below). Unfortunately, I realised I wasn't handling the centre coordinates, so I'm looking into using affine transforms now. The downside of this is that affine matrixes are handled fairly inconsistently (e.g. active vs passive) between different Python libraries, but I'm hopeful that I can get it to work once I understand what Napari expects as I have some experience with this.

Screenshot 2025-07-09 144436

fibsem/milling/patterning/plotting.py
```diff
@@ -53,20 +56,30 @@
...
+        x0, y0 = handlebox.xdescent, handlebox.ydescent
+        width, height = handlebox.width, handlebox.height
+        p0 = orig_handle.get_children()[0]
+        fc= p0.get_facecolor()
+        ec = p0.get_edgecolor()
+        patch = mpatches.Rectangle((x0, y0), width, height, facecolor=fc,
+                                   edgecolor=ec,
+                                   transform=handlebox.get_transform())
+        handlebox.add_artist(patch)
+        return patch
```
patrickcleeve2, line 70: This seems to break the legend, no patterns are drawn. 
matplotlib                           3.9.3


```bash
TypeError                                 Traceback (most recent call last)
Cell In[1], [line 19](vscode-notebook-cell:?execution_count=1&line=19)
     15 milling_stage3 = FibsemMillingStage(pattern=TrenchPattern(fillet=1.0e-6), name="Trench Pattern")
     17 image = FibsemImage.generate_blank_image(hfw=70e-6, resolution=(3072, 2048), random=True)
---> [19](vscode-notebook-cell:?execution_count=1&line=19) fig = draw_milling_patterns(image, milling_stages=[milling_stage, milling_stage2, milling_stage3], show_current=True)
     20 plt.show()

File ~/github/fibsem/fibsem/milling/patterning/plotting.py:523, in draw_milling_patterns(image, milling_stages, crosshair, scalebar, title, show_current, show_preset, show_depth, highlight_overlaps)
    521         handles.append(pc)
    522         labels.append(pc_label)
--> [523](https://file+.vscode-resource.vscode-cdn.net/home/patrick/github/fibsem/scripts/~/github/fibsem/fibsem/milling/patterning/plotting.py:523) ax.legend(
    524     handles=handles,
    525     labels=labels,
    526     handler_map={PatchCollection: _PatchCollectionHandler},
    527 )
    529 # set axis limits
    530 ax.set_xlim(0, image.data.shape[1])

File ~/miniconda3/envs/fibsem3.11/lib/python3.11/site-packages/matplotlib/axes/_axes.py:337, in Axes.legend(self, *args, **kwargs)
    220 """
    221 Place a legend on the Axes.
    222 
   (...)    334 .. plot:: gallery/text_labels_and_annotations/legend.py
    335 """
    336 handles, labels, kwargs = mlegend._parse_legend_args([self], *args, **kwargs)
--> [337](https://file+.vscode-resource.vscode-cdn.net/home/patrick/github/fibsem/scripts/~/miniconda3/envs/fibsem3.11/lib/python3.11/site-packages/matplotlib/axes/_axes.py:337) self.legend_ = mlegend.Legend(self, handles, labels, **kwargs)
    338 self.legend_._remove_method = self._remove_legend
    339 return self.legend_

File ~/miniconda3/envs/fibsem3.11/lib/python3.11/site-packages/matplotlib/legend.py:549, in Legend.__init__(self, parent, handles, labels, loc, numpoints, markerscale, markerfirst, reverse, scatterpoints, scatteryoffsets, prop, fontsize, labelcolor, borderpad, labelspacing, handlelength, handleheight, handletextpad, borderaxespad, columnspacing, ncols, mode, fancybox, shadow, title, title_fontsize, framealpha, edgecolor, facecolor, bbox_to_anchor, bbox_transform, frameon, handler_map, title_fontproperties, alignment, ncol, draggable)
    546 self._alignment = alignment
    548 # init with null renderer
--> [549](https://file+.vscode-resource.vscode-cdn.net/home/patrick/github/fibsem/scripts/~/miniconda3/envs/fibsem3.11/lib/python3.11/site-packages/matplotlib/legend.py:549) self._init_legend_box(handles, labels, markerfirst)
    551 # Set legend location
    552 self.set_loc(loc)

File ~/miniconda3/envs/fibsem3.11/lib/python3.11/site-packages/matplotlib/legend.py:896, in Legend._init_legend_box(self, handles, labels, markerfirst)
    893         text_list.append(textbox._text)
    894         # Create the artist for the legend which represents the
    895         # original artist/handle.
--> [896](https://file+.vscode-resource.vscode-cdn.net/home/patrick/github/fibsem/scripts/~/miniconda3/envs/fibsem3.11/lib/python3.11/site-packages/matplotlib/legend.py:896)         handle_list.append(handler.legend_artist(self, orig_handle,
    897                                                  fontsize, handlebox))
    898         handles_and_labels.append((handlebox, textbox))
    900 columnbox = []

TypeError: _PatchCollectionHandler.legend_artist() missing 1 required positional argument: 'handlebox'

```

patrickcleeve2 (changes requested):
Are you able to send me example bitmap files (or add one to repo) so I can test the paths/conversion.

fibsem/milling/patterning/plotting.py
thomasmfish: Ah, my bad, I completely forgot to go back to that after getting stuck in with the napari stuff - I'll work on that for a bit now. I'm now sure where this plotting is actually used though? Is it for diagnostics?

thomasmfish: Okay, I have that working now! In the example below I have plotted both a trench and a slightly taller trench bitmap pattern. In this case, the bitmap has the dwell set to full but is cycling the blanking on and off, as is shown by the black regions. I was surprised that `_detect_pattern_overlaps` was returning an empty list for this but perhaps I'm not understanding what that's for.

example pattern plotting mpl

thomasmfish:
> There are some issues with the plotting code. I've tested in the ui, and it seems alright.
> 
>     * Have you run this on the microscope?
> 
>     * Are you able to send me example bitmap files (or add one to repo) so I can test the paths/conversion.

We haven't run this on the microscope. I am doubtful that I'll be able to get spare time on there for setup and testing until we're fully on to bitmap milling, I'm currently trying to prepare for that, so that I can start testing in-silico, but I suspect we're a month or two away from properly trying bitmap adaptive milling.

Sure! Here are some [example_bitmaps](https://github.com/user-attachments/files/21218894/example_bitmaps.zip) that I used for testing how bitmaps are handled, with a summary plots file showing what's in them. In the plots the top shows the dwell modifier and blanking is showed as black in the bottom. The bitmaps are all 1D, which isn't typical of what would be used as AutoScript doesn't do any interpolation (see image below) but it matches how Thermo plots the bitmaps.

Dwell_cycling_with_zeros_after
The pattern is shown at the top, having been moved upwards. The dashed line that was milled was in the middle of the pattern in Y. If you make it 2D by adding an extra layer of pixels, it's just two lines.

fibsem/milling/patterning/plotting.py
```diff
@@ -1,11 +1,13 @@
...
 from dataclasses import dataclass
-from typing import Callable, List, Tuple
+from typing import List, Tuple, Union
 
+import numpy as np
 import matplotlib.patches as mpatches
 import matplotlib.pyplot as plt
-import numpy as np
+from matplotlib.collections import PatchCollection
+from skimage.transform import resize
```
patrickcleeve2, line 11: Missing imports
```suggestion
import logging
import math
from dataclasses import dataclass
from typing import List, Tuple, Union, Sequence

import numpy as np
import matplotlib.patches as mpatches
import matplotlib.pyplot as plt
from matplotlib.collections import PatchCollection
from matplotlib.legend_handler import HandlerBase
from skimage.transform import resize

```

fibsem/milling/patterning/plotting.py
```diff
@@ -530,11 +697,97 @@ def draw_rectangle_shape(pattern_settings: FibsemRectangleSettings, image_shape:
...
+    if pattern_settings.flip_y:
+        image_bmp = image_bmp.transpose(Image.Transpose.FLIP_TOP_BOTTOM)
+
+    image_resized = image_bmp.resize((w, h))
+
+    if not np.isclose(rotation, 0):
+        image_rotated = image_resized.rotate(-pattern_settings.rotation, expand=True)
+
+    # Create base rectangle shape
+    shape = np.asarray(image_rotated, dtype=np.float_)
```
patrickcleeve2, line 772: Doesn't work when rotation = 0
```suggestion
    if not np.isclose(rotation, 0):
        image_resized = image_resized.rotate(-pattern_settings.rotation, expand=True)

    # Create base rectangle shape
    shape = np.asarray(image_resized, dtype=np.float_)
```

patrickcleeve2 (changes requested):
Some missing imports and variables. Otherwise plotting seems to work

Image

Image

fibsem/milling/patterning/plotting.py
patrickcleeve2: Plots are for debugging, but also used for reports so it's important they work at least representative. The highlight_overlap is a newish experimental display. It was returning nothing because the bitmap mask func had a bug, that errored out. The actual overlaps aren't correct because it seems that the entire shape returns non-zero.

thomasmfish: Ah, thanks for catching that bug! I also noticed a bug that I'd dealt with in the napari drawing function where `.squeeze()` would cause issues with 1D bitmaps, so I've changed that to `.squeeze(-1)` to just get rid of the channel axis.

It seems to me that the whole bitmap should be considered non-zero (unless blanked maybe, but that may just be complicating things), as it does mill slightly even if the dwell time modifier is 0. Are you saying that the entire array is counted as non-zero, including non-bitmap areas that are due to rotations? If so, do you think it'd work to use the affine transform stuff I added to for the napari drawing? That way `draw_bitmap_shape` could be done, then turned into a boolean array before applying the transform, which should solve any issues with dwell time 0s vs padding 0s.

thomasmfish:
@patrickcleeve2 Looks like I might be able to get a bit of time to test bitmaps on 01/08, which is a cryo-cycling day, so I'll be able to let you know more after that. I'm sure you'd be welcome to join if you'd like to but it'll be from 9am UK time, so I'm guessing you'd prefer to not work super late.

Is there anything specific I should look out for or try? My plan is to create some test bitmaps and check that they are viewed and run same as if they'd been set up through the TFS GUI.

patrickcleeve2:
> @patrickcleeve2 Looks like I might be able to get a bit of time to test bitmaps on 01/08, which is a cryo-cycling day, so I'll be able to let you know more after that. I'm sure you'd be welcome to join if you'd like to but it'll be from 9am UK time, so I'm guessing you'd prefer to not work super late.
> 
> Is there anything specific I should look out for or try? My plan is to create some test bitmaps and check that they are viewed and run same as if they'd been set up through the TFS GUI.

Primary testing would be:
1. Bitmap milling and display works as expected.
2. The changes to the milling display doesn't regress or change the behaviour of any of the pattern. 

I've tested no.2 on the simulator, but need to validate it on hardware too.

fibsem/ui/napari/patterns.py
```diff
@@ -286,77 +373,138 @@ def draw_milling_patterns_in_napari(
...
+                    viewer.add_image(
+                        data=shape.shape,
+                        name=layer_name,
+                        colormap=COLOURMAPS[shape.colour],
+                        opacity=opacity,
+                        blending=blending,
+                        depiction="plane",
+                        affine=affine,
+                    )
+                layer_names_used.add(layer_name)
```
patrickcleeve2, line 495: Sorry I misunderstood this, we can't draw each stage as a separate shapes because it's too slow. You can just add this change to collect the shapes
```suggestion
    if all_napari_patterns:
        opacity = SHAPES_LAYER_PROPERTIES["opacity"]
        blending = SHAPES_LAYER_PROPERTIES["blending"]
        edge_width = SHAPES_LAYER_PROPERTIES["edge_width"]
        shapes_list: List[np.ndarray] = []
        shape_types: List[str] = []
        shape_colours: list[str] = []
        for i, (layer_name, patterns) in enumerate(all_napari_patterns.items()):
            image_list: List[NapariPattern] = []
            for pattern in patterns:
                if pattern.shape_type in IMAGE_PATTERN_TYPES:
                    image_list.append(pattern)
                else:
                    shapes_list.append(pattern.shape)
                    shape_types.append(pattern.shape_type)
                    shape_colours.append(pattern.colour)

            for shape in image_list:
                # Napari applies translate before affine, which causes issues
                # with centring for the rotation and scaling. Applying
                # translate via the affine avoids this issue.
                translate_affine = np.asarray(
                    [[1, 0, shape.translate[0]], [0, 1, shape.translate[1]], [0, 0, 1]]
                )
                affine = translate_affine @ shape.affine
                # Requires a separate layer per-image
                layer_name = f"{shape.name} {shape.shape_type} {shape.index}"
                if layer_name in viewer.layers:
                    # Update layer if it already exists
                    viewer.layers[layer_name].data = shape.shape
                    viewer.layers[layer_name].colormap = COLOURMAPS[shape.colour]
                    viewer.layers[layer_name].opacity = opacity
                    viewer.layers[layer_name].blending = blending
                    viewer.layers[layer_name].affine = affine
                else:
                    viewer.add_image(
                        data=shape.shape,
                        name=layer_name,
                        colormap=COLOURMAPS[shape.colour],
                        opacity=opacity,
                        blending=blending,
                        depiction="plane",
                        affine=affine,
                    )
                layer_names_used.add(layer_name)

        if shapes_list:
            layer_name = "Milling Patterns"
            if layer_name in viewer.layers:
                # need to clear data before updating, to account for different shapes.
                viewer.layers[layer_name].data = []
                viewer.layers[layer_name].data = shapes_list
                viewer.layers[layer_name].shape_type = shape_types
                viewer.layers[layer_name].edge_color = shape_colours
                viewer.layers[layer_name].face_color = shape_colours
                viewer.layers[layer_name].translate = translation
                viewer.layers[layer_name].opacity = opacity
                viewer.layers[layer_name].blending = blending
            else:
                viewer.add_shapes(
                    data=shapes_list,
                    name=layer_name,
                    shape_type=shape_types,
                    edge_width=edge_width,
                    edge_color=shape_colours,
                    face_color=shape_colours,
                    opacity=opacity,
                    blending=blending,
                    translate=translation,
                )
            layer_names_used.add(layer_name)
```

patrickcleeve2 (changes requested):

fibsem/ui/napari/patterns.py
thomasmfish: Ah really? That's a shame. I didn't notice any slowdown but I wasn't specifically looking for it.

My reasons for doing it this way were:
1. It makes the other patterns more consistent with bitmaps.
2. I was thinking it could be useful to be able to toggle specific stages on and off in the viewer.

thomasmfish:
Finally got around to splitting off some changes into other branches, so I'll have to rebase this branch after those PRs.

patrickcleeve2:
> > Primary testing would be:
> > ```
> > 1. Bitmap milling and display works as expected.
> > 
> > 2. The changes to the milling display doesn't regress or change the behaviour of any of the pattern.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I've tested no.2 on the simulator, but need to validate it on hardware too.
> 
> What would I do for 2? Try loading each pattern and see how they display, or would I need to try milling with each of them too?
> 
> To keep you up to date, the testing day has now been moved to 31/07.

Yeah, I would also do a quick full runthrough with on grid milling (you don't need to let it finish, just check each pattern is drawn correctly).

thomasmfish:
Hi @patrickcleeve2, I've been using the matplotlib bitmap plotting and I found it scaled pretty badly for larger bitmaps, so I've changed to the much faster `ax.imshow`. However, to keep things neat I ended up changing some of the other functions a bit, so you might want to check to see if you're happy with that.

Also, I'm testing this branch on the microscope tomorrow, so I'll let you know how that goes.

thomasmfish:
Just split off the ArrayPattern commit, I'm happy for this to be merged whenever suits you.

PR #17, opened 2025-07-28
FibsemImage.load: get filename and path from TIFF

thomasmfish:
This would be useful when using the simulated microscope


thomasmfish:
Actually, I think this probably isn't the correct way of handling it, so never mind.

PR #18, opened 2025-08-05
ArrayPattern: fix rotation

thomasmfish:
ArrayPattern was using degrees where it should have been radians.


thomasmfish:
Fixed by 3f24e6621c18521c1a6093310bc0c30a428296a3

PR #19, opened 2025-08-06
Add bitmap interpolation/resizing option

thomasmfish:
This is one way to avoid issues with bitmap patterns having an effective beam overlap that is too big or small (even negative, i.e. unconnected points), leading milling to not be as expected. This approach resizes bitmaps to ensure they have the correct pixel size, based on the pitch of a regular pattern. It does so by first drawing a rectangle, reading the pitch, then disabling it.

I'm not sure whether this will be possible for other microscopes, but it seems to me like this will most accurately match user expectations, while giving the option to turn off interpolation entirely.

### DIFF
--- a/fibsem/microscope.py
+++ b/fibsem/microscope.py
@@ -120,6 +120,7 @@ from fibsem.structures import (
 
 if TYPE_CHECKING:
     from collections.abc import Callable
+    from numpy.typing import NDArray
     from fibsem.structures import TFibsemPatternSettings
 
 class FibsemMicroscope(ABC):


fibsem/microscope.py
```diff
@@ -2619,6 +2627,60 @@ def draw_bitmap_pattern(self, pattern_settings: FibsemBitmapSettings):
...
+            center_x=pattern_settings.centre_x,
+            center_y=pattern_settings.centre_y,
+            width=pattern_settings.width,
+            height=pattern_settings.height,
+            depth=pattern_settings.depth,
+        )
+        pitch = rectangle.pitch_x
+        rectangle.enabled = False
+
+        new_shape = np.round(
```
patrickcleeve2, line 2651: Any particular reason to cast this as an np.array rather than just Tuple[int, int]?

fibsem/microscope.py
```diff
@@ -2562,10 +2564,16 @@ def draw_bitmap_pattern(self, pattern_settings: FibsemBitmapSettings):
 
         # Get bitmap from pattern settings
         bitmap_pattern = BitmapPatternDefinition()
-        bitmap_pattern.points = pattern_settings.bitmap
+
+        if pattern_settings.interpolate is not None:
```
patrickcleeve2, line 2568: You probably want to also explicitly set the application file to "Si" so that the bitmap and rectangle are drawn with the same application file. I'd do it before this interpolate function is called. Have a look at the draw_circle func for example. 
It "should" be already set via setup milling, but not guaranteed.

fibsem/structures.py
```diff
@@ -1097,6 +1097,7 @@ class FibsemBitmapSettings(FibsemPatternSettings):
     path: InitVar[Optional[Union[str, os.PathLike]]] = None
     array: InitVar[Optional[NDArray[Any]]] = None
     bitmap: Optional[NDArray[Any]] = field(init=False)
+    interpolate: Optional[Literal["nearest", "bicubic", "bilinear"]] = None
```
patrickcleeve2, line 1100: Ideally, this would be an Enum but guessing it's to avoid the gui complexity that comes with that right?

fibsem/microscope.py
thomasmfish: I don't understand this, why does it have to be Si? I don't understand why the default application file is overridden for circle in that way. Given that `self._default_application_file` defaults to "Si" but would be updated by `setup_milling` (so will always be set), wouldn't it be better manually set the default application file to `self._default_application_file`?

Looking into this, I've noticed a few potential issues with how the application file is set, so I'll make a PR based on what I think makes sense (I might be completely wrong but it should be clearer to explain).

fibsem/structures.py
thomasmfish: It's mostly to avoid adding an enum that's used in one place for one argument. Since FibsemBitmapSettings is only used internally, unlike `BitmapPattern`, it's easy enough to ensure the correct string is passed (using Literal means type checking will complain if not).

It would be good to simplify the `interpolate` and `bicubic` arguments in `BitmapPattern` and `TrenchBitmapPattern` to create a drop down list based on the typing (from either a Literal or Enum) at some point in the future though, but I didn't think it was worth creating another object until the GUI stuff was figured out.

fibsem/microscope.py
thomasmfish: Actually, I won't mess with a PR now but I'll create an issue.

patrickcleeve2: It might not have to be "Si" but it should be a application file that supports Bitmap and Rectangle, otherwise it will throw an error. And if you use a different application file for the Rectangle than Bitmap, it might have different parameters (overlap, pitch etc)

thomasmfish: I'll hold off on making a change until we've discussed #50 a bit more, since the wrapper I'm suggesting would be able to handle this using something like "Si" as a fallback. If `set_default_application_file` is called at the beginning of the `draw_bitmap` function, it should apply to both automatically, right? Or are there any issues there? I don't entirely trust the API to work as it says.

fibsem/microscope.py
```diff
@@ -2866,6 +2883,61 @@ def draw_bitmap_pattern(self, pattern_settings: FibsemBitmapSettings):
...
+            depth=pattern_settings.depth,
+        )
+
+        new_shape = (
+            int(round(pattern_settings.height / rectangle.pitch_y)),
+            int(round(pattern_settings.width / rectangle.pitch_x)),
+        )
+
+        # Disable after calculations just in case values are cleared
+        rectangle.enabled = False
```
patrickcleeve2: Not sure if it's possible in the current version, but it would be better to remove the pattern. Just a note

fibsem/milling/patterning/patterns2.py
```diff
@@ -96,7 +96,11 @@ class BitmapPattern(BasePattern[FibsemBitmapSettings]):
     scan_direction: str = "TopToBottom"
     path: str = ""
     array: Optional[NDArray[Any]] = None
-    _advanced_attributes = ("time",)
+    interpolate: bool = True
```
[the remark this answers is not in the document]
thomasmfish: Does the AutoLamella GUI now support this?

patrickcleeve2 (approved):

PR #82, opened 2026-03-12
Add configurable Tescan preset restoration, reduce default FIB imaging current, and refactor beam wait logic

ilyushkin:
Improves preset handling in the Tescan microscope driver by preserving imaging parameters during preset changes, refactoring beam readiness checks, and reducing the default FIB imaging current to 10 pA.

**Changes:**

1. **Improved preset activation**

* Added `_activate_preset()` to preserve rotation, field of view, and image shift
* Preservation is controlled by `TESCAN_PRESERVE_SETTINGS_ON_PRESET_CHANGE` and defaults to enabled

2. **Refactored beam readiness checks**

* Added `_wait_for_beam_ready()` to remove duplicated waiting logic
* Reused it in beam preparation and preset activation

3. **Safer default FIB preset**

* Changed the post-milling preset from `30 keV; 150 pA` to `30 keV; 10 pA`
* Uses `set_preset()` for consistency with parameter preservation

### DIFF
--- a/fibsem/microscope.py
+++ b/fibsem/microscope.py
@@ -554,9 +554,9 @@ class FibsemMicroscope(ABC):
         logging.debug({"msg": "get", "key": key, "beam_type": beam_name, "value": value})
         return value
 
-    def set(self, key: str, value: Union[str, float, int, tuple, list, Point], beam_type: Optional[BeamType] = None) -> None:
+    def set(self, key: str, value: Union[str, float, int, tuple, list, Point], beam_type: Optional[BeamType] = None, **kwargs) -> None:
         """Set wrapper for logging"""
-        self._set(key, value, beam_type)
+        self._set(key, value, beam_type, **kwargs)
         beam_name = "None" if beam_type is None else beam_type.name
         logging.debug({"msg": "set", "key": key, "beam_type": beam_name, "value": value})
 
@@ -565,7 +565,7 @@ class FibsemMicroscope(ABC):
         pass
 
     @abstractmethod
-    def _set(self, key: str, value: Union[str, float, int, list, tuple, Point], beam_type: Optional[BeamType] = None) -> None:
+    def _set(self, key: str, value: Union[str, float, int, list, tuple, Point], beam_type: Optional[BeamType] = None, **kwargs) -> None:
         pass
 
     # TODO: i dont think this is needed, you set the beam settings and detector settings separately
--- a/fibsem/microscope.py
+++ b/fibsem/microscope.py
@@ -1047,9 +1047,9 @@ class FibsemMicroscope(ABC):
         self.set("detector_brightness", brightness, beam_type)
         return self.get("detector_brightness", beam_type)
 
-    def set_preset(self, preset: str, beam_type: BeamType) -> str:
+    def set_preset(self, preset: str, beam_type: BeamType, preserve_settings=True) -> str:
         """Set the preset for the specified beam type."""
-        self.set("preset", preset, beam_type)
+        self.set("preset", preset, beam_type, preserve_settings=preserve_settings)
         return self.get("preset", beam_type)
 
     def _get_compucentric_rotation_offset(self) -> FibsemStagePosition:
--- a/fibsem/microscope.py
+++ b/fibsem/microscope.py
@@ -554,9 +554,9 @@ class FibsemMicroscope(ABC):
         logging.debug({"msg": "get", "key": key, "beam_type": beam_name, "value": value})
         return value
 
-    def set(self, key: str, value: Union[str, float, int, tuple, list, Point], beam_type: Optional[BeamType] = None, **kwargs) -> None:
+    def set(self, key: str, value: Union[str, float, int, tuple, list, Point], beam_type: Optional[BeamType] = None) -> None:
         """Set wrapper for logging"""
-        self._set(key, value, beam_type, **kwargs)
+        self._set(key, value, beam_type)
         beam_name = "None" if beam_type is None else beam_type.name
         logging.debug({"msg": "set", "key": key, "beam_type": beam_name, "value": value})
 
@@ -565,7 +565,7 @@ class FibsemMicroscope(ABC):
         pass
 
     @abstractmethod
-    def _set(self, key: str, value: Union[str, float, int, list, tuple, Point], beam_type: Optional[BeamType] = None, **kwargs) -> None:
+    def _set(self, key: str, value: Union[str, float, int, list, tuple, Point], beam_type: Optional[BeamType] = None) -> None:
         pass
 
     # TODO: i dont think this is needed, you set the beam settings and detector settings separately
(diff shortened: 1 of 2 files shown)


patrickcleeve2 (changes requested):
Looks pretty good, just a couple minor things

patrickcleeve2:
hi @ilyushkin are you still working on this?

fibsem/microscope.py
```diff
@@ -565,7 +565,7 @@ def _get(self, key: str, beam_type: Optional[BeamType] = None) -> Union[float, i
         pass
 
     @abstractmethod
-    def _set(self, key: str, value: Union[str, float, int, list, tuple, Point], beam_type: Optional[BeamType] = None) -> None:
+    def _set(self, key: str, value: Union[str, float, int, list, tuple, Point], beam_type: Optional[BeamType] = None, **kwargs) -> None:
```
patrickcleeve2: I don't think it's a good idea to add kwargs to this to only support the preserve settings flag for tescan presets. Also, the current code breaks the other drivers

I think it would be better if we converted it to an internal flag on the TescanMicroscope, rather than passing it as arg to the set preset functions. My understanding is that this would typically always be set true, and not something that is changed per-call?

fibsem/microscopes/tescan.py
```diff
@@ -1528,8 +1532,80 @@ def _get(self, key: str, beam_type: Optional[BeamType] = None) -> Union[float,
...
 
-    def _set(self, key: str, value, beam_type: BeamType = None) -> None:
-        """Set a property of the microscope."""
+    def _activate_preset(self, beam: Union['Automation.SEM', 'Automation.FIB'], beam_type: BeamType, preset_name: str, preserve_settings: bool) -> None:
+        """Activate a preset and optionally preserve beam settings."""
+        # Check if preset is available
+        if not beam.Preset.IsAvailable(preset_name):
+            logging.warning(f"Preset {preset_name} not available for {beam_type}.")
+            return
+
```
patrickcleeve2: **Suggestion (3 of 4): make `preserve_settings` a Tescan-local instance flag.**

This reads a flag defined by two additions on unchanged lines (can't be inline suggestions, so pasted here):

**1. Module constant** — next to `TESCAN_BEAM_READY_TIMEOUT` (~line 17):

```python
TESCAN_PRESERVE_SETTINGS_ON_PRESET_CHANGE = True   # restore rotation/FOV/shift across preset changes; False lets the preset own the optics
```

**2. Instance flag** — in `__init__`, alongside `milling_channel` / `stage_is_compustage`:

```python
self._preserve_settings_on_preset_change: bool = TESCAN_PRESERVE_SETTINGS_ON_PRESET_CHANGE
```

Then this method drops its `preserve_settings` parameter and reads the flag:

```suggestion
    def _activate_preset(self, beam: Union['Automation.SEM', 'Automation.FIB'], beam_type: BeamType, preset_name: str) -> None:
        """Activate a preset, preserving rotation/FOV/shift when self._preserve_settings_on_preset_change is set."""
        # Check if preset is available
        if not beam.Preset.IsAvailable(preset_name):
            logging.warning(f"Preset {preset_name} not available for {beam_type}.")
            return

        preserve_settings = self._preserve_settings_on_preset_change
```

fibsem/microscopes/tescan.py
```diff
@@ -1624,12 +1700,9 @@ def _set(self, key: str, value, beam_type: BeamType = None) -> None:
...
-            # check if value is in available
-            if not beam.Preset.IsAvailable(value):
-                logging.warning(f"Preset {value} not available for {beam_type}.")
-                return
-            beam.Preset.Activate(value)
-            logging.info(f"Preset {value} activated for {beam_type}.")
+            # Get preserve_settings from kwargs, defaults to True
+            preserve_settings = kwargs.get("preserve_settings", True)
+            self._activate_preset(beam, beam_type, value, preserve_settings)
             return
```
patrickcleeve2, line 1685: **Suggestion (4 of 4): no kwargs to read now — call the helper directly.**

```suggestion
        if key == "preset":
            self._activate_preset(beam, beam_type, value)
            return
```

To fully close the cross-driver break you flagged above, this also needs Tescan's `_set` signature to drop `**kwargs`, and `set()` / abstract `_set()` / `set_preset()` in `fibsem/microscope.py` reverted to their pre-PR signatures — the instance flag replaces all of that plumbing.

patrickcleeve2 (changes requested):
I'd like to change this to an internal flag for tescan, rather than adjust the public api. I don't believe you'd want to change this per call, and only set and forget it. Correct me if I'm wrong

fibsem/microscope.py
ilyushkin: Given the current implementation, I would indeed expect this option to be set to true in most cases.

However, the current implementation is a workaround. A better approach would be to update the Essence presets programmatically so that the properties we currently save here, such as scan rotation and FoV, are harmonized. The drawback is that this would modify users' presets in Essence, which some users might not appreciate, unless we created a dedicated set of presets specifically for fibsemOS/Odemis.

However, regardless of whether we continue restoring selected settings as we do now or update the presets themselves, the main issue is the beam shift. At present, we carry the beam shift from one preset to another in the hope that, when switching presets, it keeps us closer to the area of interest, even when the change in beam current is large, for example from 10 pA to 1 nA.

This approach works. Otherwise, each time we switch presets, the XY beam shift is reset to the default values stored in the preset configuration in Essence, which are not necessarily (0, 0).

fibsem/microscopes/tescan.py
```diff
@@ -1128,8 +1113,8 @@ def finish_milling(self, imaging_current: float = None, imaging_voltage: float =
             imaging_current (float): The current to use for imaging in amps.
         # """
         try:
-            default_preset = "30 keV; 150 pA"
-            self.connection.FIB.Preset.Activate(default_preset) # TODO: restore the default preset?
+            default_preset = "30 keV; 10 pA"
```
patrickcleeve2, line 1116: I will make a follow up to this to snapshot the preset during ```setup_milling```, and restore it here rather than having it hardcode.

patrickcleeve2 (approved):

fibsem/microscope.py
patrickcleeve2: At the moment I dont think it's possible update presets via the api, or read their parameters. Agree this is the right fix

ilyushkin: The `tescanautomation` module has an `Update()` method. It saves the currently modified properties, such as the scan rotation, to the specified preset. I tried it, and it works just fine. But this means that we modify the presets, so it is more invasive that the current implementation.

image

Example:
```
from tescanautomation import Automation

client = Automation("192.168.56.101")
beam = client.FIB

beam.Preset.Activate("30 keV; 10 pA")
beam.Optics.SetImageRotation(180)
beam.Preset.Update("30 keV; 10 pA")    # Update this preset to store the new rotation
beam.Preset.Activate("30 keV; 1 nA")   # Go back to 1 nA with the default 0° rotation
beam.Preset.Activate("30 keV; 10 pA")  # Activate 10 pA with the new 180° rotation
```

PR #100, opened 2026-06-05
[fix] check control type when creating controls

thomasmfish:
Should mean empty fields are created correctly.

Currently the AP field for model path is not being created in the GUI because the default is "", which is then read as None so is not created. I've also included a similar check for boolean fields to match how it's done in pattern_settings_widget.py.

### DIFF
--- a/fibsem/ui/widgets/milling_stage_editor_widget.py
+++ b/fibsem/ui/widgets/milling_stage_editor_widget.py
@@ -314,6 +314,7 @@ class FibsemMillingStageWidget(QWidget):
             if conf.get("hidden", False) is True:
                 continue
             # QUERY: use a data structure for field metadata instead of dict?
+            type_ = conf.get("type")
             label_text = conf.get("label") or name.replace("_", " ").title()
             scale = conf.get("scale")
             units = conf.get("unit")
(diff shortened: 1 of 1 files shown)


thomasmfish:
@patrickcleeve2 sorry I forgot to tag you in this!

patrickcleeve2:
Hey Thomas, thanks for the fix — the root cause diagnosis is correct. However, `milling_stage_editor_widget.py` and `milling_task_config_widget.py` are deprecated and no longer used in production (everything has migrated to `MillingTaskConfigWidget2` / `MillingTaskViewerWidget`). I'm going to remove these files now things are tested

The active equivalents that need this fix are:

- **`pattern_settings_widget.py` ~line 173** — already has `type_ is bool`, but is **missing the `type_ is str` / filepath check** for empty-string defaults. The filepath case is handled via `m.get("filepath")` but a bare `str` field with `value = ""` would fall through to `continue` (unsupported type).

- **`strategy_settings_widget.py` ~line 105** — same situation: `type_ is bool` is there, but `type_ is str` is not, so a strategy config field typed as `str` with a default of `""` won't get a control created.

Could you retarget the fix to those two files instead? The pattern is already cleaner there (no `isinstance(value, str)` fallback at all), so the fix would just be adding an `elif type_ is str:` branch before the `else: continue` in each.

PR #111, opened 2026-06-26
[ui] Migrate AutoLamella UI + FibsemUI off napari onto a matplotlib quad-view canvas

patrickcleeve2:
## Summary

Migrates the **AutoLamella main UI** and the standalone **`FibsemUI`** off napari onto a reusable
matplotlib **quad-view canvas** (`FibsemImageCanvas`) driven by a declarative **canvas-state reducer**
(`MicroscopeViewController` over `SceneModel` / `CanvasState`). Every microscope-display surface is
ported, the `FEATURE_QUAD_VIEW_ENABLED` flag is retired, and **napari is no longer imported by the
AutoLamella main window, `FibsemUI`, or the minimap widget**. A set of **main-tab UX features**
(selected view, full screen, hotkeys, working-distance scroll) is then built on the new canvas.

This started as a behind-a-flag parallel migration (phases 0–6); the PR now contains the **full cutover**.

## What's migrated

**Main tab (quad view)** — 2×2 SEM | FIB / FM | placeholder, one `FibsemImageCanvas` per beam:
- image routing, double-click-to-move, Alt/Shift canvas modifiers, drag-to-measure ruler;
- overlays via the reducer — milling patterns, alignment area, POI, spot burn, detection (mask +
  feature points), per-canvas info bar;
- **active-overlay input model** — at most one overlay per canvas owns input and suppresses
  stage-move / milling-menu / select while active (restores napari's active-layer gating).

**Fluorescence** — `FMCanvasWidget` composites per-channel frames into one RGB (additive, napari-matched
colours) with a dark layers panel (visibility / colormap / opacity / gamma / contrast), a z-slider +
max-projection toggle, double-click relative move, and Shift+scroll objective focus.

**Lamella / Protocol editor** — migrated off its `lamella_viewer` onto the canvas + reducer (reference
images, patterns, POI, editable alignment, per-task fluorescence + spot-burn).

**Spot burn** — unified onto a shared `SpotBurnSettings` coordinate editor, threaded off napari's
`thread_worker`.

**Standalone `FibsemUI`** — runs fully viewer-less on the quad-view controller; the image / movement /
FM / manipulator / objective-control widgets are all viewer-optional and napari-free.

**Overview / minimap tab** — `FibsemMinimapWidget` off its napari viewer onto
`FibsemImageCanvas`: composited overview, click-to-navigate (select nearest / move stage / add-or-move
context menu), **entity-grouped overlays** (LamellaMarkers / CurrentPosition / ReferenceFrame) on one
[...]

PR #119, opened 2026-07-06
AutoLamella shortcut creator

thomasmfish:
I'm currently writing docs for AM and I thought it would be nice to have a simple way to start AutoLamella, so I threw this together. A couple of things I'm not sure about:
1. Is there a logo included somewhere I can add?
2. Having it as `pyshortcuts` in a separate dependency group makes it a bit clunky to use but also it doesn't really belong anywhere else (the most relevant group would be 'ui' but that seems like a push).
3. I had some issues with running `uv sync`, which I can do in a separate PR for if you think it's relevant:
  - ~~I had to limit `requires-python = ">=3.8, <3.14"`~~ Turns out this is fixed by the other change
  - I had to add `"pyqt5-qt5 == 5.15.2 ; platform_system == 'Windows'"` to the 'ui' requirement group, since more recent versions haven't got a Windows version.

@patrickcleeve2  What do you think?

### DIFF
--- a/fibsem/applications/autolamella/tools/shortcut_creator.py
+++ b/fibsem/applications/autolamella/tools/shortcut_creator.py
@@ -1,0 +1,37 @@
+from __future__ import annotations
+import shutil
+
+import pyshortcuts
+
+
+def _check_create_shortcut(shortcut_type: str) -> bool:
+    msg = f"Do you wish to create a {shortcut_type} shortcut'? [Y/N]"
+    option = str(input(msg)).strip().lower()
+    if option in ("y", "yes"):
+        return True
+    elif option in ("n", "no"):
+        return False
+    print(f"Invalid option '{option}'")
+    return _check_create_shortcut(shortcut_type=shortcut_type)
+
+
+def _create_shortcut_file(
+    desktop: bool = True, start_menu: bool = True, gui: bool = True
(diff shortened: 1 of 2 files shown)


patrickcleeve2:
I think this is on the right path, but there are a few things we need to do first. The primary one is refactoring napari out of the application ([#111](https://github.com/fibsem-os/fibsem-os/pull/111) ). It overrides any icon displayed for the program, and makes bundling a bit complicated.

After that, I'd like to add a proper Windows installer that would include the desktop and start menu shortcuts. In the meantime, I don't want to add a dependency that will just be removed.

At the moment we've been suggesting people use a .bat script and add it as a shortcut, we should just document that better. If you're up for it, the most useful version of this is a short docs section with the .bat snippet (conda + uv variants) plus the make-a-shortcut step (right-click → Send to → Desktop), rather than a tool + dependency. Happy to hand over the snippets I've been using.

Separately; yes please, split the uv sync fix (the pyqt5-qt5 == 5.15.2 ; platform_system == 'Windows' pin) into its own PR. But is this a uv, qt or windows issue?

thomasmfish:
@patrickcleeve2 I've created #122 to replace this PR and I'll make a separate PR now for the pyqt5-qt5 fix

PR #123, opened 2026-07-07
Fix pyqt5-qt5 dependency on Windows

thomasmfish:
Only Linux and MacOS have pyqt5-qt5 releases > 5.12.2, so without updating to PyQt6 pyqt5-qt5 has to be set at 5.15.2 for Windows. This issue was highlighted to me by `uv` trying and failing to generate a lock file with consistent versions across the board.


thomasmfish:
@patrickcleeve2 Here's the pyqt5-qt5 fix — hopefully I've explained the issue well enough in the description.

patrickcleeve2:
Looks good, but can we remove the uv.lock file for now. Package doesn't currently support it (yet)

patrickcleeve2 (approved):

thomasmfish:
> Looks good, but can we remove the uv.lock file for now. Package doesn't currently support it (yet)

Okay, removed `uv.lock`!

PR #125, opened 2026-07-08
[docs] Recommend Miniforge and document uv install option

### LINKED ISSUE #121: Python distribution recs
Hi @patrickcleeve2, in INSTALLATION.md you're recommending people install anaconda but that can be a licensing issue for a lot of academic institutes, so it might be worth changing this to [miniforge](https://conda-forge.org/download/), which would be a drop in replacement without those issues.

patrickcleeve2:
Closes #121.

Updates `INSTALLATION.md` per the issue:

- Recommends [Miniforge](https://conda-forge.org/download/) over the Anaconda distribution — conda-forge-based, lightweight, and avoids Anaconda's channel licensing terms that affect many academic institutions.
- Adds an **Installation with uv** section documenting `uv pip install` as a faster drop-in for pip (addresses the follow-up comment), including how to install uv and the `.[ui]` extra. No lockfile or migration to `uv sync` — pip/conda flows are unchanged.

Docs-only; no dependency or tooling changes.

Both `uv pip install -e .` and `uv pip install -e '.[ui]'` were tested locally on macOS (resolve + editable install + imports OK).

### DIFF
--- a/INSTALLATION.md
+++ b/INSTALLATION.md
@@ -8,8 +8,12 @@
 
 ### Python
 Python 3.9+ is required.
-The [Anaconda distribution](https://www.anaconda.com/distribution/)
-of python is recommended.
+We recommend installing Python via [Miniforge](https://conda-forge.org/download/),
+a minimal installer that uses the community-maintained
+[conda-forge](https://conda-forge.org/) channel by default. It provides the same
+`conda` (and `mamba`) commands as Anaconda/Miniconda while staying lightweight and
+avoiding the licensing terms that can apply to Anaconda's default channels for larger
+organizations.
 
 ## Setting up your python virtual environment
 It is also highly recommended to use virtual environments for development,
@@ -17,7 +21,8 @@ see [Managing Conda Environments](https://docs.conda.io/projects/conda/en/latest
 for more information.
 (Optionally, you could use `virtualenv` if you prefer.)
 
-Create a new virutal environment from the Anaconda Prompt terminal:
+Create a new virtual environment from your conda terminal (the Miniforge Prompt on
+Windows, or any terminal on macOS/Linux):
 ```
 cd fibsem
 conda env create -n fibsem python=3.11 pip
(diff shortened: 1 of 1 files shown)


INSTALLATION.md
```diff
@@ -43,6 +48,34 @@ Once activated, move to the fibsem root directory and install fibsem like so
...
+```
+# Windows (PowerShell)
+powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"
+```
+Alternatively, install it into an existing environment with `pip install uv`.
+
+uv works inside either environment above (conda or virtualenv) with nothing extra to
+set up. Once your environment is created and activated, install fibsem with:
+```
+uv pip install -e .
```
thomasmfish: Why an editable install?

INSTALLATION.md
thomasmfish:
```suggestion
python -m pip install -e .
```

I think `python -m pip` is the correct way to call it

INSTALLATION.md
```diff
@@ -43,6 +48,34 @@ Once activated, move to the fibsem root directory and install fibsem like so
...
+```
+To include the GUI dependencies, install the `ui` extra:
+```
+uv pip install -e '.[ui]'
```
thomasmfish: I think it would make more sense to put this first and make the other option "To install without the GUI dependencies", as I assume most people will want the GUI.

INSTALLATION.md
```diff
@@ -43,6 +48,34 @@ Once activated, move to the fibsem root directory and install fibsem like so
 pip install -e .
```
thomasmfish:
```suggestion
python -m pip install -e .
```

As above

thomasmfish (changes requested):
I think this needs some info about how to set up a virtual environment with uv too. I've recently updated the installation instructions [here](https://github.com/rosalindfranklininstitute/adaptive_milling#installation), which I think makes it pretty clear.

INSTALLATION.md
patrickcleeve2: Yeah — it's a dev install from source, so editable is what you'd want when working from a GitHub checkout (tweak configs/code, no reinstall needed).

INSTALLATION.md
```diff
@@ -81,7 +85,7 @@ extra to set up — just activate that environment instead.
 
 Once your environment is active, install fibsem with the GUI dependencies (recommended):
 ```
-uv pip install -e '.[ui]'
+uv pip install -e ".[ui]"
```
thomasmfish, line 88: I don't think you need the quotation marks here, unless that's something that's different on other OSs?

thomasmfish (approved):
LGTM, just the one comment about the quotation marks

INSTALLATION.md
patrickcleeve2: they're needed for zsh on a mac, and don't do anything on other systems

PR #126, opened 2026-07-09
[feat] Add separate width per trench component

ilyushkin:
This pull request enhances the flexibility of the trench pattern generation in `fibsem/milling/patterning/patterns2.py`.
This is required for waffle milling since the trench components can have not only different height but also different width.

The main improvements are the addition of independent upper and lower trench width parameters, and a more robust calculation of fillet geometry to ensure proper fitting within trench dimensions.

**Pattern parameter enhancements:**

* Added `upper_trench_width` and `lower_trench_width` as optional parameters to `TrenchPattern`, allowing the upper and lower trenches to have different widths. If not specified, these default to the main `width` value.

**Trench and fillet geometry improvements:**

* Updated the `define` method so that `upper_trench_width` and `lower_trench_width` are used for their respective trenches, improving control over trench shape.
* Improved the calculation of fillet size: the fillet radius is now constrained by both trench height and width, preventing geometry errors. The code now computes separate inner widths for upper and lower trenches to position fillets and fills accurately.
* Adjusted the placement of fillet and fill shapes to use the correct inner widths and positions, ensuring that all shapes fit within the specified trench dimensions.

### DIFF
--- a/fibsem/milling/patterning/patterns2.py
+++ b/fibsem/milling/patterning/patterns2.py
@@ -597,6 +597,24 @@ class TrenchPattern(BasePattern[Union[FibsemRectangleSettings, FibsemCircleSetti
             "tooltip": "Height of the lower trench.",
         },
     )
+    upper_trench_width: Optional[float] = field(
+        default=None,
+        metadata={
+            "label": "Upper Trench Width",
+            **DEFAULT_DISTANCE_METADATA,
+            "advanced": True,
+            "tooltip": "Width of the upper trench. If None, uses Width.",
+        },
+    )
+    lower_trench_width: Optional[float] = field(
+        default=None,
+        metadata={
+            "label": "Lower Trench Width",
+            **DEFAULT_DISTANCE_METADATA,
+            "advanced": True,
+            "tooltip": "Width of the lower trench. If None, uses Width.",
+        },
+    )
     cross_section: CrossSectionPattern = field(
         default=CrossSectionPattern.Rectangle, metadata=DEFAULT_CROSS_SECTION_METADATA
     )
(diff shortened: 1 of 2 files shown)


fibsem/milling/patterning/patterns2.py
```diff
@@ -597,6 +597,24 @@ class TrenchPattern(BasePattern[Union[FibsemRectangleSettings, FibsemCircleSetti
             "tooltip": "Height of the lower trench.",
         },
     )
+    upper_trench_width: Optional[float] = field(
```
patrickcleeve2, line 600: These render as nothing in the milling pattern editor. `pattern_settings_widget._build_controls` picks the widget by `isinstance(value, ...)`, and `None` matches no branch → it hits the `else`, logs `"Control for '…' is unsupported"`, and drops the field. So the params can't be set from the GUI (two spurious warnings per rebuild), and a spinbox has no way to represent `None` even if it did render.

Since `width` is the interface we want to keep, suggest collapsing to one field with a `0` sentinel — plain float, renders fine, and `0` is a safe out-of-band value (a zero-width trench is meaningless):

```python
lower_trench_width: float = field(
    default=0.0,
    metadata={
        "label": "Lower Trench Width",
        **DEFAULT_DISTANCE_METADATA,
        "minimum": 0.0,   # override: allow the 0 sentinel (default min is 0.01)
        "advanced": True,
        "tooltip": "Width of the lower trench. Set to 0 to match Width.",
    },
)
```

`width` stays the upper/reference width; lower falls back to it unless explicitly set. Backward compatible (old protocols → `0.0` → identical geometry).

fibsem/milling/patterning/patterns2.py
```diff
@@ -660,7 +672,7 @@ def define(self) -> List[Union[FibsemRectangleSettings, FibsemCircleSettings]]:
         )
 
         upper_trench_settings = FibsemRectangleSettings(
-            width=width,
+            width=upper_trench_width,
```
patrickcleeve2, line 675: The main rects now keep full width, which removes the rounded corners the fillet is meant to produce. Because milling only ever removes material, a full-width rectangle over the corner can't be "rounded back" by the circle afterward — it mills more than the old narrowed-rect + fillet path did.

Restore the per-trench narrowing when `fillet > 0`:

```python
upper_trench_width = self.width
lower_trench_width = self.lower_trench_width if self.lower_trench_width > 0 else self.width

max_fillet = min(upper_trench_height/2, lower_trench_height/2,
                 upper_trench_width/2, lower_trench_width/2)
fillet = float(np.clip(self.fillet, 0.0, max_fillet))

upper_main_width = upper_trench_width - 2*fillet if fillet > 0 else upper_trench_width
lower_main_width = lower_trench_width - 2*fillet if fillet > 0 else lower_trench_width
# ...use *_main_width for the two main rects, *_inner widths for fillet/fill placement
```

Worth a quick on-hardware check too, since the fillet offset is empirically tuned — but geometrically the full-width rect defeats it. The `test_define_with_fillet` / `test_fillet_clipping` assertions would revert to the narrowed widths.

patrickcleeve2 (commented):
Thanks for adding per-trench widths — the waffle use case makes sense. Two things to sort before merge: one milling-geometry regression and one that makes the new params unusable from the GUI. Details inline.

**1. Fillet no longer rounds the corners.** The main trench rectangles are now drawn at full width instead of being narrowed by `2·fillet`. A full-width rectangle mills straight through the corner, so it removes strictly more than the old fillet path — the tuned rounded corner gets squared off, and the fillet circles/fills end up sitting inside the rectangle doing nothing. This changes the default trench-with-fillet output, and the updated `test_define_with_fillet` (asserting `width == 100` instead of `90`) bakes the regression in. Suggest restoring the per-trench narrowing when `fillet > 0`.

**2. The new `Optional[float] = None` fields don't work in the pattern editor.** `pattern_settings_widget` dispatches controls on the *runtime value type*, so a `None` value matches nothing, falls through to the `else`, logs `"Control for 'upper_trench_width' is unsupported"`, and skips the field — the params are invisible/uneditable and you get two warnings on every trench-editor rebuild. And even if it rendered, a spinbox can't represent `None`, so "unset → use width" can't round-trip. Suggest a single `lower_trench_width: float = 0.0` sentinel (`0 = match width`) instead.

fibsem/milling/patterning/patterns2.py
patrickcleeve2: image

Issue #127, opened 2026-07-10
Parameter 'passes' does not show up in pattern configuration GUI


thomasmfish:
We want to be able to change the dwell time for patterns, and the only way to do that in fibsemOS seems to be changing the number of passes but that parameter doesn't appear in the GUI. Also, according to the metadata the max passes is set to 100, which is way below the multiple millions of passes we are often doing. I've found some other GUI bugs that I'm working on fixes for but I am not sure why this is happening.

PR #128, opened 2026-07-10
Strategy and pattern GUI control fixes

thomasmfish:
Hi @patrickcleeve2 

Some bits of AutoLamella currently don't handle integers properly, which means that a strategy can receive a `float` when it's expecting an `int`, for example (this caused some issues for a session recently). I've made a separate `IntegerValueSpinBox` and split up some of the GUI logic to handle this.

I also noticed that strategies with `QFilePathLineEdit` weren't having the values updated from the GUI, so I've added that too. To prevent stuff like that going unnoticed, I've also added some errors that will come up if there are controls that aren't being read, which should make it easier to catch such issues going forward.

I also noticed that passes aren't being shown as an option in the GUI, so I've created issue #127 for that.

### DIFF
--- a/fibsem/ui/widgets/custom_widgets.py
+++ b/fibsem/ui/widgets/custom_widgets.py
@@ -14,6 +14,7 @@ from PyQt5.QtWidgets import (
     QAction,
     QComboBox,
     QDoubleSpinBox,
+    QSpinBox,
     QFileDialog,
     QHBoxLayout,
     QLabel,
@@ -234,6 +235,35 @@ class ValueComboBox(QComboBox):
     def value(self):
         """Return the raw value stored as item data for the current selection."""
         return self.currentData()
+
+
+class IntegerValueSpinBox(QSpinBox):
+    """QSpinBox with sensible defaults, WheelBlocker, and None-safe configuration."""
+
+    def __init__(
+        self,
+        suffix: Optional[str] = None,
+        minimum: Optional[int] = None,
+        maximum: Optional[int] = None,
+        step: Optional[int] = None,
+        tooltip: Optional[str] = None,
+        no_buttons: bool = False,
+        parent=None,
+    ) -> None:
+        super().__init__(parent)
+        if suffix:
+            self.setSuffix(f" {suffix}")
+        self.setRange(
+            minimum if minimum is not None else 0,
+            maximum if maximum is not None else 1000000,
+        )
+        self.setSingleStep(step if step is not None else 1)
+        if tooltip:
+            self.setToolTip(tooltip)
+        if no_buttons:
+            self.setButtonSymbols(QAbstractSpinBox.ButtonSymbols.NoButtons)
+        self.setKeyboardTracking(False)
+        self.installEventFilter(WheelBlocker(parent=self))
 
 
 class ValueSpinBox(QDoubleSpinBox):
--- a/fibsem/ui/widgets/pattern_settings_widget.py
+++ b/fibsem/ui/widgets/pattern_settings_widget.py
@@ -27,6 +27,7 @@ from fibsem.ui.widgets.custom_widgets import (
     QFilePathLineEdit,
     ValueComboBox,
     ValueSpinBox,
+    IntegerValueSpinBox,
 )
 
 _HIDDEN_FIELDS = {"shapes"}
@@ -161,6 +162,9 @@ class FibsemPatternSettingsWidget(QWidget):
             dims = m.get("dimensions")
             effective_scale = (base_scale ** dims) if (base_scale and dims) else base_scale
 
+            if effective_scale is None:
+                effective_scale = 1
+
             if items == "dynamic":
                 cached = self.microscope.get_available_values_cached(
                     m["microscope_parameter"], BeamType.ION
@@ -179,22 +183,21 @@ class FibsemPatternSettingsWidget(QWidget):
                 control = QCheckBox()
                 control.setChecked(bool(value))
             elif type_ is int:
+                effective_scale = int(round(effective_scale))
                 suffix = (
                     utils._get_display_unit(base_scale, m.get("unit"))
                     if base_scale
                     else (m.get("unit") or "")
                 )
-                # Ensure integer types don't have a step size under 1 and have 0 decimals
-                control = ValueSpinBox(
+                # Ensure integer types don't have a step size under the effective scale or 1
+                step = max(m.get("step", 1), effective_scale)
+                control = IntegerValueSpinBox(
                     suffix,
                     m.get("minimum"),
                     m.get("maximum"),
-                    max(m.get("step", 1), 1),
-                    0,
+                    step,
                 )
-                control.setValue(
-                    int(round(value * effective_scale if effective_scale else value))
-                )
+                control.setValue(int(round(value * effective_scale)))
             elif isinstance(value, (float, int)) or type_ is float:
                 suffix = (
                     utils._get_display_unit(base_scale, m.get("unit"))
@@ -208,7 +211,7 @@ class FibsemPatternSettingsWidget(QWidget):
                     m.get("step"),
                     m.get("decimals"),
                 )
-                control.setValue(value * effective_scale if effective_scale else value)
+                control.setValue(value * effective_scale)
             else:
                 logging.warning("Control for '%s' is unsupported", field_name)
                 continue  # unsupported type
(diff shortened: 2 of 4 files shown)


fibsem/ui/widgets/custom_widgets.py
```diff
@@ -236,6 +237,35 @@ def value(self):
         return self.currentData()
 
 
+class IntegerValueSpinBox(QSpinBox):
+    """QSpinBox with sensible defaults, WheelBlocker, and None-safe configuration."""
+
+    def __init__(
+        self,
+        suffix: Optional[str] = None,
+        minimum: Optional[float] = None,
```
patrickcleeve2: these should be Optional[int] type?

fibsem/ui/widgets/pattern_settings_widget.py
```diff
@@ -178,17 +182,36 @@ def _build_controls(self, pattern: BasePattern) -> None:
...
+                )
+                # Ensure integer types don't have a step size under the effective scale or 1
+                step = max(m.get("step", 1), effective_scale)
+                control = IntegerValueSpinBox(
+                    suffix,
+                    m.get("minimum"),
+                    m.get("maximum"),
+                    step,
+                )
+                control.setValue(int(round(value * effective_scale)))
```
patrickcleeve2, line 200: is it possible to do this type coercion inside the class, rather than requiring to do it outside.

fibsem/ui/widgets/pattern_settings_widget.py
```diff
@@ -265,7 +291,7 @@ def get_pattern(self) -> BasePattern:
                 x = row.x_control.value() / _POINT_SCALE
                 y = row.y_control.value() / _POINT_SCALE
                 pattern.point = Point(x=x, y=y)
-            elif isinstance(row.control, ValueComboBox):
+            elif isinstance(row.control, (ValueComboBox, IntegerValueSpinBox)):
```
patrickcleeve2: this is grouped with combobox, should be with ValueSpinBox below

fibsem/ui/widgets/pattern_settings_widget.py
```diff
@@ -308,10 +338,18 @@ def set_pattern(self, pattern: BasePattern) -> None:
                 row.control.set_value(value)
             elif isinstance(row.control, ValueSpinBox):
                 row.control.setValue(value * row.scale if row.scale else value)
+            elif isinstance(row.control, IntegerValueSpinBox):
+                row.control.setValue(
+                    int(round(value * row.scale if row.scale else value))
+                )
```
patrickcleeve2, line 351: consolidating the type coercion inside the class, would also help with this duplication

fibsem/ui/widgets/strategy_settings_widget.py
```diff
@@ -118,16 +126,36 @@ def _build_controls(self, strategy: MillingStrategy[Any]) -> None:
...
-            elif isinstance(value, (float, int)):
+            elif type_ is int:
+                effective_scale = int(round(effective_scale))
+                suffix = (
+                    utils._get_display_unit(base_scale, m.get("unit"))
+                    if base_scale
+                    else (m.get("unit") or "")
+                )
+                # Ensure integer types don't have a step size under the effective scale or 1
+                step = max(m.get("step", 1), effective_scale)
```
patrickcleeve2, line 137: why is this clipped to the effective scale, not sure I understand?

fibsem/ui/widgets/strategy_settings_widget.py
```diff
@@ -202,9 +238,24 @@ def get_strategy(self) -> MillingStrategy[Any]:
...
-                setattr(strategy.config, row.field, val / row.scale if row.scale else val)
+                setattr(
+                    strategy.config, row.field, val / row.scale if row.scale else val
+                )
+            elif isinstance(row.control, IntegerValueSpinBox):
+                val = row.control.value()
+                setattr(
+                    strategy.config,
+                    row.field,
+                    int(round(val / row.scale if row.scale else val)),
```
patrickcleeve2, line 249: same comment here about type coercion

patrickcleeve2 (changes requested):

fibsem/ui/widgets/strategy_settings_widget.py
```diff
@@ -143,18 +171,26 @@ def _build_controls(self, strategy: MillingStrategy[Any]) -> None:
...
-            elif isinstance(control, ValueSpinBox):
+            elif isinstance(control, (ValueSpinBox, IntegerValueSpinBox)):
                 control.valueChanged.connect(self._on_changed)
             elif isinstance(control, QCheckBox):
                 control.toggled.connect(self._on_changed)
+            elif isinstance(control, QFilePathLineEdit):
+                control.editingFinished.connect(self._on_changed)
+            else:
+                raise TypeError(
+                    f"Unsupported control type '{type(control)}' in FibsemPatternSettingsWidget"
```
patrickcleeve2: should be FibsemStrategySettingsWidget

fibsem/ui/widgets/strategy_settings_widget.py
thomasmfish: Based on my understanding of how your UI works, if the scale is 10 we don't want steps of 1 because that couldn't be represented as an integer, so at a minimum it must be steps of 10. Does that make sense?

fibsem/ui/widgets/pattern_settings_widget.py
thomasmfish: It makes sense to me that an integer control should throw an error if given something other than an integer, so I personally don't like the idea of it doing this conversion automatically. I could make it so that `setValue` has an argument `round: bool = False` that can be set to True in these cases but I think this seems simpler, honestly.

fibsem/ui/widgets/pattern_settings_widget.py
```diff
@@ -269,13 +295,17 @@ def get_pattern(self) -> BasePattern:
                 data = row.control.value()
                 if data is not None:
                     setattr(pattern, row.field, data)
-            elif isinstance(row.control, ValueSpinBox):
+            elif isinstance(row.control, (ValueSpinBox, IntegerValueSpinBox)):
```
patrickcleeve2: This re-introduces the float bug for integer fields. `value()` returns an `int`, but `val / row.scale` is true division, so `5 / 1 == 5.0` — the field is stored as a float, defeating the point of `IntegerValueSpinBox`. (`row.scale` is already an int here, so enforcing an int scale wouldn't help; it's the `/`.)

Split it out like `get_strategy` does in the strategy widget:

```python
elif isinstance(row.control, ValueSpinBox):
    val = row.control.value()
    setattr(pattern, row.field, val / row.scale if row.scale else val)
elif isinstance(row.control, IntegerValueSpinBox):
    val = row.control.value()
    setattr(pattern, row.field, int(round(val / row.scale if row.scale else val)))
```

This also restores symmetry — strategy splits the two, pattern currently combines them.

patrickcleeve2:
Could we add a round-trip test (`set_*` → `get_*`) asserting an integer field like `passes`/`n_columns` comes back as `int`? This is the second time this float-leak has surfaced, so a regression test would lock it down.

I've drafted one (offscreen Qt, no pytest-qt dependency, mirrors `tests/fm/test_autofocus_widget.py`) — it fails on the current head (`n_columns` comes back as `7.0`) and passes with the `get_pattern` split suggested above. Happy to open it as a follow-up.

fibsem/ui/widgets/pattern_settings_widget.py
patrickcleeve2: ok fair. qt has the same type checking strictness, so all good

patrickcleeve2:
@thomasmfish — I put a regression test + the `get_pattern` fix (from the inline comment above) on a branch you can pull straight into this PR, if useful. Two commits:

- `7afd3789` — test: guard integer field round-trip in the pattern settings widget (fails on this branch's head — `n_columns` comes back as `7.0`)
- `8cd6a305` — fix: give `IntegerValueSpinBox` its own branch in `get_pattern` with `int(round(...))` (makes it pass)

To cherry-pick both onto `fix-integer-controls`:

```
git fetch https://github.com/fibsem-os/fibsem-os.git claude/integer-control-tests
git cherry-pick 7afd3789 8cd6a305
```

Or take just the test and apply the fix yourself. The test is headless (offscreen Qt, no `pytest-qt`), mirroring `tests/fm/test_autofocus_widget.py`:

```
QT_QPA_PLATFORM=offscreen pytest tests/ui/test_settings_widgets_integer.py
```

No pressure on how it lands — happy for you to own the fix.

fibsem/ui/widgets/pattern_settings_widget.py
thomasmfish: Yep, sorry about that! I got muddled

thomasmfish:
Sorry, I didn't see that you'd made the fix before fixing it myself, but I've cherry picked your tests and neatened up the commits a bit

patrickcleeve2 (approved):

PR #130, opened 2026-07-10
Increase maximum passes to 100 million

thomasmfish:
@patrickcleeve2 to deal with the limit mentioned in #127 

We've observed upwards of 2.5 million passes for a pattern, so 100 million seems like a generous upper limit while allowing for these numbers.

### DIFF
--- a/fibsem/milling/properties.py
+++ b/fibsem/milling/properties.py
@@ -54,7 +54,7 @@ DEFAULT_PASSES_METADATA = {
     "label": "Passes",
     "type": int,
     "minimum": 0,
-    "maximum": 100,
+    "maximum": 100000000,
     "step": 1,
     "advanced": True,
     "tooltip": "Number of passes for the pattern. Set to 0 for automatic calculation." 


thomasmfish:
Hmm, setting the passes doesn't seem to actually work (at least from the test today), has it worked for you in the past and are there any limitations?

patrickcleeve2:
> Hmm, setting the passes doesn't seem to actually work (at least from the test today), has it worked for you in the past and are there any limitations?

It probably won't work as expected, there is a circular dependency between passes, dwell-time, depth, and 'pass-count' (https://github.com/fibsem-os/fibsem-os/blob/945b2235b60bbb9ece1506aafa4956df6a1c7b4f/fibsem/microscope.py#L3066)

Only RegularCrossSectionPattern applies the passes directly. What type of pattern were you using?

thomasmfish:
We were testing with Bitmap patterns, so maybe that was the problem. We were basically finding that the bitmap milling is much slower than expecting, so we wanted to see whether changing the number of passes/dwell time would help. It throws an error if you try to set a value outside of some internally specified range but then if you read the value after setting it doesn't seem to have changed, which is a bit confusing. Do you know more about what the limitations are for setting passes on the API side? It seems to be possible via XT.

patrickcleeve2:
For bitmaps, I haven't really tried that out. But for rectangles, i remember it being difficult via the api. Setting passes directly would alter the depth, rather than maintain the depth and alter the dwell time. 
There is definitely a difference between doing it via api than xt.

patrickcleeve2 (approved):

PR #138, opened 2026-07-15
Spot burn: status-bar progress, unified UI, and milling-style workflow hook (FIB-249)

patrickcleeve2:
Implements [FIB-249](https://linear.app/fibsemos/issue/FIB-249). Follow-up to the spot-burn crash fix (#137). Three related commits:

## 1. Status-bar progress + unified progress/button styling

- Add `microscope.spot_burn_progress_signal`; `run_spot_burn` emits through it and no longer takes `parent_ui` (fully UI-decoupled).
- **Unsupervised** runs now show progress in the status-bar `FibsemProgressWidget`; **supervised** runs show it in both the widget and the status bar, from one signal via a shared mapping helper (`Burning spots — N/M · Xs remaining` → `Done`).
- The spot burn widget renders through the **same** `FibsemProgressWidget` as the status bar (guaranteed-consistent formatting), disconnecting on teardown.
- Run/Cancel button uses the app-standard `PRIMARY` / `STOP_WORKFLOW` stylesheets (matching the milling widget).

## 2. Task cleanup — empty-coordinates hang fix

- The unsupervised/headless path now takes the automatic branch whenever there's no user in the loop, and **skips with a warning if no coordinates are set** instead of falling through to the interactive `ask_user`, which would block forever in an unattended run.
- Removed a stale commented-out `parent_ui` argument.

## 3. Supervised runs via a milling-style workflow hook

Supervised spot burns are now task-orchestrated, matching the milling workflow:
- The user places points and clicks a **"Run Spot Burn"** button in the workflow controls; the task runs the burn and **waits for completion** before continuing (the workflow can no longer advance mid-burn), then re-prompts so the user can re-burn (preview loop) or press **Continue**.
- Widget: `start_spot_burn_signal` (BlockingQueuedConnection) + `is_burning`; `set_workflow_mode()` hides the widget's own Burn button during the workflow (a hint points to the workflow control) but shows it as **"Cancel"** while a burn is in progress so cancellation stays available locally.
- Workflow **Stop** now cancels an in-progress burn (previously the task aborted but the beam kept burning on the worker).

## Tests

`tests/autolamella/test_spot_burn.py` covers coordinate filtering, string-param coercion, config serialization, progress emission, the empty-coords skip, and the supervised run/wait/continue loop. UI-dependent tests skip gracefully where napari/PyQt aren't installed (CI).

## Verification

Unit tests + end-to-end plumbing checks, plus a full GUI pass on the running app: manual and automated runs, cancel-burn, and cancel-workflow all confirmed.

### DIFF
--- a/fibsem/applications/autolamella/ui/AutoLamellaMainUI.py
+++ b/fibsem/applications/autolamella/ui/AutoLamellaMainUI.py
@@ -11,7 +11,7 @@ except Exception:
 import warnings
 
 import napari
-from PyQt5.QtCore import Qt
+from PyQt5.QtCore import Qt, QTimer
 from PyQt5.QtWidgets import (
     QAction,
     QApplication,
@@ -53,6 +53,7 @@ from fibsem.ui.stylesheets import (
     WORKFLOW_BORDER_STYLESHEET,
 )
 from fibsem.ui.widgets.progress_widget import FibsemProgressWidget, ProgressUpdate
+from fibsem.ui.FibsemSpotBurnWidget import build_spot_burn_progress_update
 from fibsem.ui import notification_service
 from fibsem.ui.widgets.autolamella_lamella_protocol_editor import (
     AutoLamellaProtocolEditorWidget,
@@ -881,6 +882,15 @@ class AutoLamellaSingleWindowUI(QMainWindow):
             self.autolamella_ui.microscope.tiled_acquisition_signal.connect(
                 self._on_tile_acquisition_progress
             )
+            try:
+                self.autolamella_ui.microscope.spot_burn_progress_signal.disconnect(
+                    self._on_spot_burn_progress
+                )
+            except Exception:
+                pass
+            self.autolamella_ui.microscope.spot_burn_progress_signal.connect(
+                self._on_spot_burn_progress
+            )
         self.btn_create_experiment.setEnabled(True)
         self.btn_load_experiment.setEnabled(True)
         self._update_instructions()
@@ -923,6 +933,13 @@ class AutoLamellaSingleWindowUI(QMainWindow):
 
         elif state == "finished":
             self.milling_progress_bar.setVisible(False)
+
+    @ensure_main_thread
+    def _on_spot_burn_progress(self, ddict: dict) -> None:
+        """Handle spot burn progress updates from the microscope (supervised + unsupervised)."""
+        self.progress_widget.update_progress(build_spot_burn_progress_update(ddict))
+        if ddict.get("finished"):
+            QTimer.singleShot(2000, self.progress_widget.reset)
 
     @ensure_main_thread
     def _on_tile_acquisition_progress(self, ddict: dict) -> None:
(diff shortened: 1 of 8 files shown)


patrickcleeve2:
## Review summary

Full review of the three commits at bd402f9a (progress-signal decoupling, empty-coordinates skip, milling-style supervised workflow hook). All findings below are addressed in the follow-up commit ad510dc5.

### What's solid
- The microscope-level `spot_burn_progress_signal` matches the existing `milling_progress_signal` / `tiled_acquisition_signal` pattern and fully decouples `run_spot_burn` from the UI; the disconnect-before-connect in the main window follows the established idiom.
- `build_spot_burn_progress_update` gives the widget and the status bar a single formatting source; `ProgressUpdate` handles the zero/empty edge cases safely (no division-by-zero paths).
- The supervised run/wait/re-prompt loop faithfully mirrors the milling loop in `tasks/base.py`, including the `BlockingQueuedConnection` start signal.
- Teardown hygiene: `disconnect_signals()` before `deleteLater()` on microscope reconnect; worker→GUI marshalling via `@ensure_main_thread` on both subscribers.

### Issues found → fixed in ad510dc5
1. **Workflow abort left the widget stuck in workflow mode** (medium). `set_workflow_mode(True)` is applied on every spot-burn `ask_user`, but the only reset was the `clear_spot_burn` message emitted at normal task completion — pressing **Stop** mid-task raised `InterruptedError` before it ran, leaving the Burn button hidden with a stale hint until a microscope reconnect. Fix: `_workflow_finished` now restores the widget (workflow mode + points layer), mirroring the existing milling run-button restore right above it.
2. **Stop-vs-start race could leave a burn running with no owner** (narrow window, high consequence). `run_spot_burn_worker` clears `stop_event` on entry; if Stop landed between the task's last abort check and the main thread servicing the blocking-queued slot, the burn started *after* `cancel_spot_burn()` had already fired, the task died on its first wait-loop check, and nothing cancelled the burn. Fix: the wait loop now cancels the burn on `InterruptedError` before re-raising, so an aborting task always takes its burn down with it. (Milling shares this latent pattern — `stop_milling` is a no-op when not yet milling — worth a separate look.)

### Nits → fixed in ad510dc5
- Grid positions for the progress widget and workflow hint are now derived from the layout (`getItemPosition`) instead of hardcoded; the deliberately shared button/hint cell is documented (visibility is mutually exclusive, so the hint swaps in without layout shift).
- The "BlockingQueuedConnection guarantees is_burning is already True" comments overstated the contract (the burn can be refused when no in-bounds points are placed) — reworded on both the task and widget side.
- The 2 s hide-after-done timer on the shared status-bar progress widget could hide a newer operation's progress → added `FibsemProgressWidget.reset_if_finished()`.
- An errored burn showed "Done" → added `ProgressUpdate.failed` and an `error` flag on the finished emit; both surfaces now show "Spot burn failed".

### Behavior change worth noting
Unsupervised runs with no stored coordinates previously fell through to an interactive prompt (blocking forever when headless); they now skip with a log warning. Correct semantics for "unsupervised", but the skip is log-only.

### Tests
New: supervised abort cancels the burn; progress-dict → `ProgressUpdate` mapping (running / done / failed). Suite: 17/17. Also verified with an offscreen widget smoke test (layout invariants, guarded reset, failed-state rendering) on top of the earlier full GUI pass.

PR #141, opened 2026-07-17
test+fix: integer field round-trip in pattern settings widget

patrickcleeve2:
Stacked on **#128** (`fix-integer-controls`). Base branch `pr128-integer-controls-base` is a snapshot of #128's head so this PR shows only the delta — retarget to `main` once #128 merges, then delete the base branch.

## What this adds

**Regression test** — `tests/ui/test_settings_widgets_integer.py`

Guards that metadata-declared `int` fields round-trip through the settings widgets as Python `int`, not `float` — the exact class of bug #128 exists to fix. Builds `FibsemPatternSettingsWidget` for an `ArrayPattern` (`n_columns` / `n_rows` / `passes`) and asserts:

- int fields get an `IntegerValueSpinBox` control
- `get_pattern()` returns them as `int`
- a `set_pattern()` → `get_pattern()` round-trip preserves `int`
- float fields (`width` / `height`) stay `float` (no over-correction)

Headless, offscreen Qt, no `pytest-qt` dependency — mirrors `tests/fm/test_autofocus_widget.py`.

**Fix** — `get_pattern()` in the pattern settings widget

On #128's head, `get_pattern` groups `IntegerValueSpinBox` with `ValueSpinBox`:

```python
setattr(pattern, row.field, val / row.scale if row.scale else val)
```

`val` is an `int`, but `/` is true division, so `5 / 1 == 5.0` — integer fields leak back as floats (even at scale 1), re-introducing the bug. Split into its own branch with `int(round(...))`, mirroring `get_strategy` in the strategy widget.

The two are separate commits, so the fix can be dropped if you'd rather it land directly in #128 (see the inline review comment there).

## Verification

Run with the project env + offscreen Qt:

```
QT_QPA_PLATFORM=offscreen pytest tests/ui/test_settings_widgets_integer.py
```

- Test commit alone (on #128 head): 2 fail — `n_columns` comes back as `7.0`
- With the fix commit: 4 pass

### DIFF
--- a/fibsem/ui/widgets/pattern_settings_widget.py
+++ b/fibsem/ui/widgets/pattern_settings_widget.py
@@ -295,9 +295,16 @@ class FibsemPatternSettingsWidget(QWidget):
                 data = row.control.value()
                 if data is not None:
                     setattr(pattern, row.field, data)
-            elif isinstance(row.control, (ValueSpinBox, IntegerValueSpinBox)):
+            elif isinstance(row.control, ValueSpinBox):
                 val = row.control.value()
                 setattr(pattern, row.field, val / row.scale if row.scale else val)
+            elif isinstance(row.control, IntegerValueSpinBox):
+                val = row.control.value()
+                setattr(
+                    pattern,
+                    row.field,
+                    int(round(val / row.scale if row.scale else val)),
+                )
             elif isinstance(row.control, QCheckBox):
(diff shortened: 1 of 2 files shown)


patrickcleeve2:
Closing in favour of handing these two commits to the author to fold into #128 (see comment there) — no separate PR needed.

PR #183, opened 2026-07-24
[fix] qtawesome complaint at scios startup about spinner object

CraignRush:
Starting fibsem-os on a TFS Scios leads to a qtawesome error where a spinner animation is calling stop before it was started. This prevents any startup of the application. A check before trying to stop the spinner animation resolves the issue.

### DIFF
--- a/fibsem/ui/widgets/custom_widgets.py
+++ b/fibsem/ui/widgets/custom_widgets.py
@@ -643,7 +643,8 @@ class _SpinnerLabel(QLabel):
 
     def stop(self):
         self._active = False
-        self._spin.stop()
+        if self._spin is not None and self._spin.parent_widget in self._spin.info:
+            self._spin.stop()
         self.update()        # repaint blank
 
     def clear(self):


fibsem/ui/widgets/custom_widgets.py
```diff
@@ -643,7 +643,8 @@ def start(self):
 
     def stop(self):
         self._active = False
-        self._spin.stop()
+        if self._spin is not None and self._spin.parent_widget in self._spin.info:
+            self._spin.stop()
         self.update()        # repaint blank
```
patrickcleeve2, line 648: Minor: `self._spin` is assigned in `__init__` and never set to `None`, so that half of the condition is dead — this keeps the same guard without the noise.

```suggestion
    def stop(self):
        self._active = False
        if self._spin.parent_widget in self._spin.info:
            self._spin.stop()
        self.update()        # repaint blank
```

patrickcleeve2 (commented):
Nice catch on the root cause — this genuinely blocks startup on older qtawesome.

Two follow-ups that aren't one-click here (so I've left them as notes rather than suggestions):

1. **`start()` has the identical bug.** [`start()`](https://github.com/fibsem-os/fibsem-os/blob/956ecaf7d45229c281f4ca75c9306130a6d15cf8/fibsem/ui/widgets/custom_widgets.py#L637-L642) calls `self._spin.start()` right after an async `self.update()`, so the widget isn't registered in `_spin.info` yet. On the same qtawesome 1.3.x that crashes here, the first acquisition will `KeyError` in `start()` the same way — worth guarding it identically.
2. **The durable fix is bumping the floor.** `pyproject.toml` pins `qtawesome>=1.3.0`, but 1.3.0 introduced `Spin.start/stop` *without* the guard; qtawesome added `if self.parent_widget in self.info` to both methods in **1.4.0** (PR spyder-ide/qtawesome#277). Bumping to `qtawesome>=1.4.0` fixes both methods upstream and is still Python 3.8-safe (qtawesome 1.4.x is `requires_python>=3.7`). Once that floor lands, these call-site guards become redundant and can be removed. Note the pin bump only reaches a machine that actually re-resolves deps — a frozen/source-checkout Scios may need a one-off `pip install -U qtawesome`.

patrickcleeve2:
Tracked in [FIB-304](https://linear.app/fibsemos/issue/FIB-304/spinner-crash-on-qtawesome-13x-guard-startstop-before-first-paint-or) — covers guarding `start()` too and the `qtawesome>=1.4.0` floor bump as follow-ups.

patrickcleeve2:
Thanks for tracking this down — the diagnosis was exactly right, and it's what made the upstream fix easy to find.

Closing in favour of #241, which bumps the floor to `qtawesome>=1.4.0`. 1.4.0 added the `if self.parent_widget in self.info` guard inside qtawesome itself (spyder-ide/qtawesome#277), so it covers `start()` as well as `stop()` — `start()` has the identical bug and would have bitten on the first acquisition. That makes the call-site guard here redundant once the pin lands.

One thing to flag for your Scios: the pin only takes effect where dependencies actually get re-resolved. If that machine is a frozen or source checkout, it'll still need a one-off

    pip install -U qtawesome

to pick up 1.4.x. Please reopen if that doesn't clear it.

PR #186, opened 2026-07-27
Export experiment targets to the automated lamella targeting ML format (FIB-305)

patrickcleeve2:
Closes [FIB-305](https://linear.app/fibsemos/issue/FIB-305/util-to-export-autolamella-experiment-data-to-targeting-ml-format).

## What

Turns the targets an operator picked during an experiment into training data for the automated lamella targeting model.

**One sample per lamella:** the final FIB reference image from its *Select Milling Position* task — the same kind of image the targeting pipeline runs on — labelled with the operator's point of interest.

## Why

The only existing producer of this format is `GridScreeningRun.to_ml_format()`, which exports the pipeline's *own* predicted masks, so training on it is self-reinforcing. Experiments are the opposite: every POI is a human-selected target on a real grid, with the outcome recorded alongside it. Every site running AutoLamella is already generating this and there was no path to use it.

## Output

Written to `<experiment>/targeting-export` by default, so a copied experiment directory carries its training data with it:

```
targeting-export/
    images/0001.tif      FIB reference image, metadata preserved
    points/0001.json     source of truth: exact coordinates + provenance
    labels/0001.tif      binary uint8 disc at the point, derived from points/
    manifest.json        index over the samples, plus skip reasons
```

`points/NNNN.json` is one flat object per lamella: sub-pixel coordinates (not rounded — that precision is the reason the sidecar exists), the original POI in milling coordinates, the milling-pose stage position, `shape` as named `{height, width}`, `pixelsize`, `hfw`, and provenance. The disc radius is given in **metres** and converted through each image's pixel size, so a label means the same physical size regardless of field of view. `images/` + `labels/` matches the pairing `upload_to_hf.create_dataset` already consumes.

## Notes for review

- **No stage reprojection.** The reference image is acquired at the milling position, so the POI — already a milling-coordinate offset in metres from the image centre — converts straight to pixels. Because it is held in metres, it lands correctly even though the operator picked it on a differently-zoomed image earlier in the task.
- **The source image is derived, not hardcoded.** `task_name` is user-configurable per protocol, so the exporter reads it off the lamella's `SELECT_MILLING_POSITION` config, then takes the highest `res_NN` of the final set. `field_of_views` sorts largest to smallest, so that is the most zoomed — and the image the task itself treats as its result.
- **Failures are kept, not dropped.** A target a human picked that then failed is a useful hard negative, so those are exported and tagged. Lamellae that never ran the task are skipped with the reason recorded in the manifest, and numbering stays contiguous across skips.
- **Writing inside the experiment directory is re-entrant.** A second export cannot ingest the first one's images (different directory and naming); there is a test for it.
- **The Development menu action does not prompt for a directory.** The destination is a subfolder that does not exist yet, which `open_existing_directory_dialog` cannot select. It exports to the default and opens the folder afterwards.

## Please check before merging

- [x] ~~Run it once against a real experiment.~~ **Done** — verified by @patrickcleeve on a real AutoLamella experiment; the derived filename patterns match what the acquisition code actually writes on disk.
- [ ] **`DEFAULT_DISC_RADIUS_M = 2.0e-6`** is a judgement call, not a measured value. Needs a domain sanity check.

## Known gap — POI may never have been set

The exporter gates on the reference image existing, not on the POI having been chosen. `select_poi_ui` returns `None` in unsupervised mode (`validate=False`) or with `select_poi=False`, in which case `Lamella.poi` keeps its template default while the final images are still acquired — so an unsupervised run would export labels at the untouched default position.

The POI can also be set from the lamella protocol editor, entirely outside the task, so "did the task ask?" is not a sound signal, and such data is rescuable by later review. Deliberately not addressed here; see FIB-305.

## Testing

34 unit tests covering source-image selection (including SEM and non-final decoys), the milling-coordinate y-flip, field-of-view independence, contiguous numbering across skips, the default path, re-entrancy, and that the label disc lands on the sidecar coordinates. Full suite green.

Also exposed as a CLI for batch conversion across experiments:

```
python -m fibsem.applications.autolamella.tools.ml_export <experiment-dir> [...]
```

### DIFF
--- a/fibsem/applications/autolamella/tools/ml_export.py
+++ b/fibsem/applications/autolamella/tools/ml_export.py
@@ -15,10 +15,10 @@ experiment carries its training data with it::
         labels/0001.tif      disc at the point, rasterised from points/
         manifest.json        index over the samples, plus experiment provenance
 
-One sample is one lamella. ``points/`` is authoritative; ``labels/`` is a pure
-derivative of it, rebuildable at any radius via :func:`regenerate_labels` without
-re-reading the source experiments, so it should never be hand-edited. The ``images/``
-+ ``labels/`` pairing is what
+One sample is one lamella. ``points/`` is authoritative and ``labels/`` is derived
+from it, so the raster should never be hand-edited -- to change the label geometry,
+re-export at a different ``radius_m`` (or rasterise the sidecar yourself). The
+``images/`` + ``labels/`` pairing is what
 :func:`~fibsem.applications.autolamella.tools.upload_to_hf.create_dataset` consumes.
 
 No stage reprojection is involved: the reference image is acquired at the milling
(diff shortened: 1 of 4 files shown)


patrickcleeve2:
**Verified against real experiments** (14 in `fibsem/applications/autolamella/log/`, 21 lamellae). Output written to a temp directory; no experiment folders were touched.

**The derived filename convention holds.** Real data writes:

```
ref_Setup Lamella Position_final_res_01_ib.tif
ref_Setup Lamella Position_final_res_02_ib.tif
ref_Spot Burn Fiducial_final_res_02_ib.tif
ref_Mill Fiducial_final_res_02_ib.tif
```

Note the task is named **"Setup Lamella Position"**, not the `SelectMillingPositionTaskConfig` display name. Deriving `task_name` from each lamella's own config is what makes this work — a hardcoded name would have matched nothing and skipped every lamella.

**Coordinates check out on operator-set POIs**, at 65.1 nm/px:

| lamella | POI (m) | expected | exported |
|---|---|---|---|
| `01-exotic-monkey` | x = 1.598e-05 | +245 px | +245 px |
| `02-clear-corgi` | y = 6.147e-06 | −94 px (y-flip) | −94 px |

**Skips and fallbacks behave.** One lamella with no reference images was skipped with the reason recorded in the manifest; numbering stayed contiguous.

**Disc radius sanity:** 2.0e-6 m at 100 µm HFW gives a 30.7 px radius, painting 0.19% of a 1024×1536 image.

Most POIs in these particular experiments sit at the untouched default — they are simulator runs where the POI was never moved, not a defect in the export.

PR #204, opened 2026-07-29
[ui] Resolve the fibsem.ui widget exports lazily (FIB-352)

patrickcleeve2:
Closes FIB-352.

## Summary

`fibsem/ui/__init__.py` imported all eight widgets at module level, and one of them
(`FibsemImageSettingsWidget`) imports napari. Because importing a submodule runs its package's
`__init__` first, reaching **any** module under `fibsem.ui` — including Qt-free leaves like
`fibsem.ui.qt.threading` — pulled in every widget, and napari with them.

| | before | after |
| --- | --- | --- |
| `import fibsem.ui` → napari / Qt loaded | True / True | **False / False** |
| `import fibsem.ui.qt.threading` → napari | True | **False** |
| `import fibsem.ui.widgets.canvas` → napari | True | **False** |

The widgets still import napari — you just stop paying for it when importing something
unrelated.

## Why this matters for the napari deprecation

A widget can be made napari-free in its own source (as #200 did for the manipulator) and, until
now, there was no way to demonstrate it: the package init dragged napari in regardless. This
makes "is this widget napari-free?" answerable by import, which is the verification story for
the rest of the cutover.

It does **not** make the UI tests run on CI — CI installs `.[test]` and the workflow says the
`[ui]` extra is omitted deliberately, so anything importing a Qt object still skips. What it
does buy is that genuinely Qt-free modules become importable there and can be tested directly.

## The one subtlety

Each widget class shares its name with the module defining it, and Python binds an imported
submodule as an attribute of its package. So once `fibsem.ui.FibsemMovementWidget` (the module)
has been imported by anything, that name is no longer *missing* — PEP 562's `__getattr__` never
runs, callers get a module, and `FibsemMovementWidget(...)` fails with "module is not callable".
The eager version papered over this by overwriting the attribute with the class at import time.

`_LazyExports` restores the guarantee by unwrapping the shadowing module on access, so the
re-exports behave exactly as they did. There is a test for precisely this ordering.

Internal callers now import from the concrete module instead — clearer, and unambiguous. Only
four sites used the re-exports.

## Side effects checked

Each in a fresh interpreter, with import order controlled:

- `DETECTION_AVAILABLE` — `False` before and after (it is computed on first access now, and is
  still a plain `bool`);
- `dir()` still advertises every export;
- an unknown attribute raises `AttributeError`, not an `ImportError` surfacing from some
  unrelated widget;
- re-exports resolve to the **class**, in both clean and shadowed import order;
- all five importable entry points still import (`FibsemUI`, `AutoLamellaMainUI`,
  `FMAcquisitionWidget`, `microscope_config_widget`, `correlation_tab_widget`);
- no star-imports of `fibsem.ui` exist anywhere to worry about.

**Not fixed, and not caused by this**: constructing the AutoLamella main window under
`QT_QPA_PLATFORM=offscreen` segfaults in vispy's GL context. Identical exit code (139) on `main`
— a pre-existing offscreen napari limitation.

## Testing

New `tests/ui/test_lazy_ui_exports.py`. Four of its six tests import nothing but `fibsem.ui`, so
they **run on CI** — the Qt skip is a decorator rather than a module-level `importorskip`, which
would skip the whole file including the part that matters there. Verified by re-running with
`napari`/`PyQt5`/`qtpy`/`superqt` blocked via a `sys.meta_path` finder: 4 passed, 2 skipped.

Full suite: 1337 passed, 15 skipped (up 6).

### DIFF
--- a/fibsem/applications/autolamella/ui/AutoLamellaMainUI.py
+++ b/fibsem/applications/autolamella/ui/AutoLamellaMainUI.py
@@ -40,7 +40,7 @@ import fibsem.config as fibsem_cfg
 from fibsem.applications.autolamella.structures import AutoLamellaTaskStatus, Lamella
 from fibsem.applications.autolamella.ui.AutoLamellaUI import AutoLamellaUI, INSTRUCTIONS
 from fibsem.applications.autolamella.workflows.tasks.tasks import get_task_supervision
-from fibsem.ui import FibsemMinimapWidget
+from fibsem.ui.FibsemMinimapWidget import FibsemMinimapWidget
 from fibsem.ui.qt.gc import install_main_thread_gc
 from fibsem.ui.stylesheets import (
     MILLING_PROGRESS_BAR_STYLESHEET,
--- a/fibsem/applications/autolamella/ui/AutoLamellaUI.py
+++ b/fibsem/applications/autolamella/ui/AutoLamellaUI.py
@@ -28,17 +28,14 @@ from fibsem.structures import (
     MicroscopeSettings,
     Point,
 )
-from fibsem.ui import (
-    DETECTION_AVAILABLE,
-    FibsemCryoDepositionWidget,
-    FibsemImageSettingsWidget,
-    FibsemMinimapWidget,
-    FibsemMovementWidget,
-    FibsemSystemSetupWidget,
-    FibsemSpotBurnWidget,
-    MillingTaskViewerWidget,
-    stylesheets,
-)
+from fibsem.ui import DETECTION_AVAILABLE, stylesheets
+from fibsem.ui.FibsemCryoDepositionWidget import FibsemCryoDepositionWidget
+from fibsem.ui.FibsemImageSettingsWidget import FibsemImageSettingsWidget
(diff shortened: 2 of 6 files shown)


patrickcleeve2:
Parking this as a draft — it is a nice-to-have, not a prerequisite for the napari removal, and it is not worth carrying the risk while #111 is the thing that actually needs attention.

To be clear about what it does and does not buy: every widget can be made napari-free without this change (verification just happens by reading source rather than by importing, which is what `tests/ui/test_gui_thread_migration.py` already does), and once napari is actually gone the eager init stops mattering by itself. What survives napari removal is smaller: leaf modules import without pulling in eight widgets and Qt, and genuinely Qt-free logic becomes testable on CI without the `[ui]` extra.

The work is complete and verified if it is ever wanted — see FIB-352 for the two non-obvious hazards it has to handle, both of which cost a round of debugging to find.

The current red CI on this branch is **not** from this change: `main` is red at cb120124 (#203), where `tests/ui/test_about_dialog.py` imports Qt at module level with no `importorskip`. Fix prepared separately.

PR #259, opened 2026-08-03
[ui] Make the supervision toggle reach the run (FIB-443)

patrickcleeve2:
Follow-up to #258. Found while wiring that dialog, but written after #258 had already merged, so it lands separately.

## The bug

The status-bar controls applied their value by writing through `_active_strategies`:

```python
def _on_supervised_toggled(self, supervised: bool) -> None:
    for strategy in self._active_strategies:
        strategy.config.supervised = supervised
```

`_connect_coincidence_strategies` only populates that list **inside** `_run_milling`. Before the first mill of a session it is empty, so the write went nowhere — and then `_seed_controls_from_strategy` seeded the button back from the untouched config, reverting the operator's choice without a word.

## Why it matters

The failure is in the **unsafe direction**. Supervised is precisely the mode where the drop trigger stops nothing:

```python
if not self.config.supervised and self._drop_detected:
    self.microscope.stop_milling()
```

So the sequence is: switch to **Automated** → button confirms it → start milling → button flips back to **Supervised**, and the run never auto-stops on the intensity drop. The operator asked for automatic stopping, was shown it was set, and doesn't get it.

It works on the *second* run of a session, because `_active_strategies` is populated by then — which would make it read as a mis-click rather than a bug.

## The fix

One owner per direction: **config → controls on load, controls → config on edit.**

The write now targets the strategies on the **live** stages. `MillingStageListWidget.get_stages()` hands out the rows' own `FibsemMillingStage` objects, so those are what a run actually uses; `get_config()` deep-copies, which is why the original code had to go via `_active_strategies` and inherited its lifetime.

Two things came along with it:

- **Seeding now runs after a config load**, which it didn't before. Without it the status bar kept the previous mode while the loaded config held a different one, and the next toggle would push the stale value back over what was just loaded. Making the toggle authoritative is what turns that into a real problem.
- **The drop-threshold control had the identical bug** — same root cause, same one-line change. Leaving one of two siblings broken would be worse than either.

## Tests

Three, reproducing the operator sequence from the issue rather than the internals. Mutation-checked by reverting each writer back to `_active_strategies`; each fails its corresponding test.

139 widget tests and 2,136 in `tests/` pass.

Closes FIB-443.

### DIFF
--- a/fibsem/ui/widgets/fluorescence_coincidence_viewer_widget.py
+++ b/fibsem/ui/widgets/fluorescence_coincidence_viewer_widget.py
@@ -1049,6 +1049,11 @@ class FluorescenceCoincidenceViewerWidget(QWidget):
                 self._apply_fm_configuration(fm_config)
             if milling_config is not None and self.milling_viewer_widget is not None:
                 self.milling_viewer_widget.set_config(milling_config)
+                # Config → controls, the load direction. Without this the status bar
+                # keeps the previous supervision mode and drop threshold while the
+                # loaded config holds different ones, and the toggle would then push
+                # the stale values back over what was just loaded.
+                self._seed_controls_from_strategy()
         except Exception as e:
             logging.error(f"Failed to apply coincidence configuration: {e}")
             QMessageBox.warning(self, "Load Configuration", f"Failed to apply:\n{e}")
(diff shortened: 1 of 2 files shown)


patrickcleeve2:
Closing — the bug this fixes cannot happen, and the change regresses the path that matters.

**FIB-443's premise is wrong.** The supervision toggle, the pause button and the drop-threshold spinbox are *run-time only* controls: hidden at construction, revealed in `_on_milling_progress` when a mill starts, hidden again when it finishes. There is no path that shows them beforehand.

```
1143:  self.btn_supervised.setVisible(False)     # construction
2091:  self.btn_supervised.setVisible(True)      # milling started
2138:  self.btn_supervised.setVisible(False)     # milling finished
```

So the sequence in the issue — open the viewer, click the toggle, start milling — requires clicking a button that is not on screen. `_active_strategies` was always the right target, precisely because these controls only exist while it is populated. The intent is stated in the source at the spinbox's construction: *"push live changes to any running strategies (not just at milling start)"*.

**And the change makes things worse.** `_run_milling` passes `get_config()`'s result to `run_milling`, and `get_settings` ends with `copy.deepcopy` — so the running strategies are copies, and are *not* reachable through the stage list. Retargeting the writes at the live stages means a mid-mill toggle reaches nothing, which is the one thing these controls are for.

Nothing here is salvageable: the config-load seeding this also added is redundant, since the controls are invisible outside a run and `_seed_controls_from_strategy` already runs when they appear.

`main` is untouched — this never merged.

PR #263, opened 2026-08-03
[hooks] Don't report an aborted workflow as completed (FIB-465)

patrickcleeve2:
Part 1 of FIB-465 — the defects only. The completion events (`EXPERIMENT_COMPLETED`, `LAMELLA_COMPLETED`) follow in a separate PR.

## The bug

`TaskManager._run_queue()`'s loop exits when the queue drains **or** when the user presses Stop, and it fired `WORKFLOW_COMPLETED` unconditionally after it:

```python
while not self.is_stopped:
    ...
self._fire_workflow_hook(HookEvent.WORKFLOW_COMPLETED)   # both ways out
update_status_ui(self.parent_ui, "", workflow_info="All tasks completed.")
```

So pressing Stop fired a completion event and put "All tasks completed." in the status bar. There was a `TASK_CANCELLED` but no workflow equivalent, so nothing downstream could tell an abort from a finished run. FIB-461 proposes generating the shareable PDF and overview PNG off `WORKFLOW_COMPLETED`, which would have regenerated them after every abort and labelled the result complete.

## Changes

**`WORKFLOW_CANCELLED`** — fired when the loop exits with the stop event set, with the status message corrected on that path.

**`TASK_SKIPPED`** — a skip was already reported to the UI with a `skip_reason`, but fired no hook, so a run that skipped every task was indistinguishable from one where every task succeeded. It fires from the manager rather than `AutoLamellaTask`, because `_should_skip` decides before any task object exists — which is also why there is no `task_state` on this event. `task_type` is looked up from the lamella's config so `Hook.task_types` filtering still works.

**`HookContext.skip_reason`** — needed to carry the reason. Rendered in message templates (`{skip_reason}`, empty string when unset) and included in the webhook payload. Growing the context rather than widening the gap with the status dict, per FIB-464.

**The silent skip path** — when `get_lamella_by_name` returned `None` the item was marked `Skipped` and dropped with no status, no hook and no log entry. It now warns and fires `TASK_SKIPPED` with reason `lamella_not_found`.

## Compatibility

Both events are additive. `HookEvent` is a `str, Enum` whose values are what `Hook.to_dict` persists, so saved configurations keep loading and firing unchanged. `hook_config_widget` derives its event list from the enum, so both are selectable in the UI without a second edit:

```
task_started, task_completed, task_failed, task_cancelled, task_skipped,
workflow_started, workflow_completed, workflow_cancelled
```

## Testing

`TaskManager` had no test coverage at all, so this adds `tests/autolamella/test_task_manager_hooks.py` — 6 tests driving `_run_queue` without running a real task (the queue drains, is stopped, or contains only items that get skipped).

Checked against the pre-fix manager: 4 of the 6 fail, and the 2 that pass are the controls (`workflow_completed` on a clean drain; a skip firing no `task_started`). 659 tests pass overall.

### DIFF
--- a/fibsem/applications/autolamella/workflows/tasks/manager.py
+++ b/fibsem/applications/autolamella/workflows/tasks/manager.py
@@ -170,7 +170,7 @@ class TaskManager:
             update_status_ui(self.parent_ui, "", workflow_info="Workflow cancelled.")
         else:
             self._fire_workflow_hook(HookEvent.WORKFLOW_COMPLETED)
-            update_status_ui(self.parent_ui, "", workflow_info="All tasks completed.")
+            update_status_ui(self.parent_ui, "", workflow_info=self._completion_message())
         print(self.experiment.task_history_dataframe())
 
     def build_run_summary_dataframe(self) -> pd.DataFrame:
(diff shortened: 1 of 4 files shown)


patrickcleeve2:
Added the aggregate closing message (6aab4cd3), since the completion string had the same failure mode as the abort bug: a run in which every task was skipped read exactly like one where all of them succeeded.

The closing line is now derived from the queue rather than fixed:

| outcome | message |
|---|---|
| nothing ran | `No tasks ran: all N skipped.` |
| some ran, some skipped | `All tasks completed (N skipped).` |
| none skipped | `All tasks completed.` (unchanged) |
| empty queue | `No tasks to run.` |
| stopped | `Workflow cancelled.` |

The all-skipped case also logs a warning, because `update_status_ui` only logs its `msg` argument and not `workflow_info` — so a headless run would otherwise get no record of a workflow that did nothing.

4 more tests, 10 in the file, 663 passing.

PR #264, opened 2026-08-03
[hooks] Add item and experiment completion events (FIB-465)

patrickcleeve2:
Part 2 of FIB-465, completing it. Part 1 (the abort defect and `TASK_SKIPPED`) shipped in #263.

Nothing could tell when a unit of work was actually *finished*. `WORKFLOW_COMPLETED` means "this run's queue drained", which is true after running one task on one lamella — so it says nothing about whether anything is done.

## Changes

**`LAMELLA_COMPLETED`** — fires the first time a lamella satisfies the workflow's completion predicate. Lead event of the two: a lamella is what gets delivered to a TEM, so the downstream recipient's unit is the lamella, and a per-lamella artifact can be produced the moment that lamella finishes rather than waiting for the whole run.

**`EXPERIMENT_COMPLETED`** — the roll-up. Same predicate over every position, fired *after* `WORKFLOW_COMPLETED` (finishing the run is what makes the experiment finishable) and **not** on the cancelled path, since an aborted run establishes nothing.

Both fire on the **transition**, not the state. The manager snapshots what was already complete when the run started, so re-running a task on finished work stays quiet.

## Three ways the predicate lies

All guarded, because FIB-461 will hang artifact generation off these events and a false positive means generating a report for work that never happened:

| case | without the guard |
|---|---|
| `required_tasks` is empty | `is_completed()` is vacuously `True`, so a protocol that failed to load announces **every** lamella as finished |
| `task_protocol` is `None` | `AttributeError` — it's `None` until a protocol is loaded ("must be set externally") |
| experiment has no lamellae | `all([])` is `True`, so an empty experiment reports as completed |

The `None` case was found by the tests, not by reading — worth noting since it's reachable from a fresh experiment.

A failed task can't trigger a false completion: `task_history` is only appended in `post_task()`, which a raising task never reaches.

## Payload

`LAMELLA_COMPLETED` carries the lamella in `item_name` plus the task that finished it. `EXPERIMENT_COMPLETED` carries the **experiment name** in `item_name` — which is what the `lamella_name` → `item_name` rename in #260 was for; the event's subject genuinely isn't a lamella.

## Compatibility

Additive, so saved hook configurations keep loading. Both events are selectable in the config UI without a second edit:

```
task_started, task_completed, task_failed, task_cancelled, task_skipped,
lamella_completed, workflow_started, workflow_completed, workflow_cancelled,
experiment_completed
```

The manual harness's LoggingHook now derives its list from `HookEvent` too, so it stops needing an edit per event.

## Testing

`tests/autolamella/test_task_manager_completion_hooks.py` — 9 tests driving completion through the *real* predicate (a workflow config with required tasks, and lamellae whose `task_history` satisfies it) rather than stubbing it, so they break if the predicate's inputs change. 719 tests pass overall.

## One naming question before merge

`LAMELLA_COMPLETED` puts AutoLamella vocabulary into `fibsem/hooks.py`, which #260 deliberately made domain-agnostic — the whole point of `item_name`. The consistent alternative is `ITEM_COMPLETED`, with `item_name` saying which lamella.

I implemented the name FIB-465 specifies rather than renaming your spec unilaterally, but **this is the moment to decide**: `HookEvent` values are persisted in hook configs, so renaming after release is a breaking change, exactly as the issue notes for its other members. Two-line change if you want `ITEM_COMPLETED`.

### DIFF
--- a/fibsem/applications/autolamella/workflows/tasks/manager.py
+++ b/fibsem/applications/autolamella/workflows/tasks/manager.py
@@ -253,8 +253,8 @@ class TaskManager:
         return bool(positions) and all(self._is_complete(p) for p in positions)
 
     def _maybe_fire_lamella_completed(self, lamella: 'Lamella', task_name: str) -> None:
-        """Fire LAMELLA_COMPLETED the first time a lamella satisfies the workflow's
-        completion predicate during this run.
+        """Fire ITEM_COMPLETED the first time a lamella satisfies the workflow's
+        completion predicate during this run. The lamella is AutoLamella's item.
 
         A lamella is the unit that gets delivered to a TEM, so this is the event
         downstream work hangs off — it lets a per-lamella artifact be produced the
@@ -265,7 +265,7 @@ class TaskManager:
         self._completed_lamella.add(lamella.name)
         fire_event(
             self.hook_manager,
-            HookEvent.LAMELLA_COMPLETED,
+            HookEvent.ITEM_COMPLETED,
             task_name=task_name,
             task_type=lamella.task_state.task_type,
             item_name=lamella.name,
@@ -275,7 +275,7 @@ class TaskManager:
     def _maybe_fire_experiment_completed(self) -> None:
         """Fire EXPERIMENT_COMPLETED once, when the last outstanding lamella finishes.
 
-        The roll-up of LAMELLA_COMPLETED: same predicate, applied to every position.
+        The roll-up of ITEM_COMPLETED: same predicate, applied to every position.
         """
         if self._experiment_was_complete or not self._all_lamella_complete():
             return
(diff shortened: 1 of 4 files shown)


patrickcleeve2:
Renamed to `ITEM_COMPLETED` / `item_completed`, settling the open question in the description — `fibsem/hooks.py` is application-agnostic and `HookContext.item_name` exists so an event's subject need not be a lamella. Done before merge, since the enum values persist in hook configs.

Two deliberate asymmetries:

- **The manager-side method keeps its lamella name** (`_maybe_fire_lamella_completed`). That layer *is* AutoLamella, where the item is a lamella; only the shared contract needs to be neutral.
- **`EXPERIMENT_COMPLETED` stays.** An experiment is a microscopy concept rather than an AutoLamella one, so it does not carry the same problem `LAMELLA_COMPLETED` did.

Event set is now:

```
task_started, task_completed, task_failed, task_cancelled, task_skipped,
item_completed, workflow_started, workflow_completed, workflow_cancelled,
experiment_completed
```

719 tests still pass.

PR #265, opened 2026-08-03
[autolamella] Record every task outcome, and register metadata outside the GUI

patrickcleeve2:
Closes FIB-490 and FIB-449.

## Failures were recorded durably nowhere (FIB-490)

`task_history` is appended in one place — `post_task` — and `run()` re-raises before reaching it:

```python
try:
    self._run()
except Exception as e:
    self._fire_hook(...)
    raise                      # <- post_task never runs
self.post_task()               # <- only on success
```

So **`task_history` contained successes only.**

The manager does set `Failed` and call `experiment.save()`, but it writes to `task_state` — a single object that the next task's `pre_task` overwrites, resetting status, message, end timestamp and outputs. The queue is task-outer / lamella-inner, so a failure survived only until that lamella reached its next task, then disappeared from `experiment.yaml` entirely. The logfile was the only remaining trace.

That silently emptied the report's Task History table, made per-protocol success rates uncomputable, and meant a bug-report bundle carried an experiment whose history showed no failures.

### What changed

`_record_outcome()` is shared by both paths and runs on every terminal outcome. On the failure path it runs **before** the hook fires, not after — previously a `task_failed` consumer read `InProgress` off `task_state`, because the manager set the real status only once the exception had propagated out of `run()`.

Two things beyond the literal fix:

**The recording is guarded.** Code in an `except` block that runs before `raise` must not itself raise, or it replaces the task's real exception with its own. There's a test asserting a deliberately broken recorder still lets `ValueError("the real failure")` through.

**The manager's cancellation predicate now matches the task's.** It was missing `InterruptedError`, so a bare one would have frozen `Cancelled` into history while the manager stamped `Failed` on the live state — the two disagreeing about the same task.

The manager's status-setting stays rather than being deleted as redundant: it's still the only thing that runs if `run_task` raises *before* the task's `run()` is entered — an unknown task name, or a construction failure.

## Registration only happened in the GUI (FIB-449)

`utils._register_metadata` was called from exactly one place, `AutoLamellaUI`. Images acquired from a script, a headless run, or the standalone FibsemUI carried the microscope's default `FibsemExperiment()` — `id=None` — and could not be associated with an experiment afterwards.

`Experiment.register_metadata(microscope)` mirrors `Experiment.configure_logging()` from FIB-421, including the deliberate omission from `load`/`create`: reading an experiment must not reach into the caller's microscope, since the load dialog reads one on every click to preview it. `TaskManager.__init__` calls it so headless runs through `run_tasks()` record it too.

## Scoped out on purpose

FIB-449's second half — stamping lamella and task identity into image metadata via the `pre_task`/`post_task` bracket. That needs the field decisions in FIB-466 and FIB-483, and building the mechanism now would leave it with no consumer. The exception-safety groundwork it depends on is what this PR lays.

## Tests

10 new tests across two files. The one that matters most asserts a history entry survives the next task's `pre_task` overwriting the live state — that's the actual bug, and a version holding a reference rather than a frozen copy would pass every other test here.

Full suite green locally (2265 passed, 24 skipped — the skips are the uninstalled plugin fixture).

### DIFF
--- a/fibsem/applications/autolamella/structures.py
+++ b/fibsem/applications/autolamella/structures.py
@@ -1369,6 +1369,32 @@ class Experiment:
         """
         return _configure_logging(path=self.path, log_filename="logfile")
 
+    def register_metadata(self, microscope: "FibsemMicroscope") -> None:
+        """Stamp this experiment's identity onto the images ``microscope`` acquires.
+
+        Which experiment produced an image is a property of the run, not of the GUI.
+        This previously happened only in AutoLamellaUI, so images acquired from a
+        script, a headless run, or the standalone FibsemUI carried the microscope's
+        default ``FibsemExperiment()`` -- id ``None`` -- and could not be associated
+        with an experiment afterwards. See FIB-449.
+
+        Mirrors ``configure_logging``: ``load`` and ``create`` deliberately do not
+        call it, because reading an experiment must not reach into the caller's
+        microscope. Callers that own the run say so. The app calls it when it adopts
+        an experiment, TaskManager calls it when it takes one to run, and a script
+        driving the microscope directly calls it itself.
+        """
+        import fibsem
+        from fibsem.utils import _register_metadata
+
+        _register_metadata(
+            microscope=microscope,
+            application_software="autolamella",
+            application_software_version=fibsem.__version__,
+            experiment_name=self.name,
+            experiment_method="null",
+        )
+
     def save_protocol(self) -> None:
         """Save the task protocol to disk in the experiment directory."""
         self.task_protocol.save(os.path.join(self.path, "protocol.yaml"))
--- a/fibsem/applications/autolamella/structures.py
+++ b/fibsem/applications/autolamella/structures.py
@@ -320,9 +320,18 @@ class AutoLamellaWorkflowConfig:
         return []
 
     def get_completed_tasks(self, lamella: 'Lamella', with_timestamps: bool = False) -> List[str]:
-        """Get the list of completed tasks for a given lamella."""
+        """Get the list of completed tasks for a given lamella.
+
+        Filtered on status for the same reason as Lamella.completed_tasks:
+        task_history holds every terminal outcome since FIB-490. Unfiltered, a
+        failed required task would make is_completed() true, which drives the
+        ITEM_COMPLETED / EXPERIMENT_COMPLETED events and the is_completed column
+        in the experiment summary.
+        """
         completed_tasks = []
         for task in lamella.task_history:
+            if task.status is not AutoLamellaTaskStatus.Completed:
+                continue
             if task.name in self.workflow:
                 txt = task.name
                 if with_timestamps:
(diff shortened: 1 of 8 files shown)


patrickcleeve2:
Good catch — confirmed, and item 1 is blocking. Pushed the fix in ea30ee95.

I wrote the regression tests before touching anything and got **5 failures against this branch**, including the gate:

```
FAILED test_a_failed_task_is_not_a_completed_task
FAILED test_a_cancelled_task_is_not_a_completed_task
FAILED test_a_failed_prerequisite_does_not_satisfy_the_gate
FAILED test_a_failed_required_task_leaves_the_workflow_incomplete
FAILED test_last_completed_task_skips_a_failed_entry
```

**The `is_failure` net is worse than "won't fire".** `is_failure` reads `defect.state`, and I checked every assignment: `defect` is only ever set in three UI widgets where a *human clicks* to mark a lamella failed, plus a legacy `from_dict` migration. Nothing sets it when a task fails automatically. So `missing_prereqs` was the only guard, and this PR removed it.

**One reader you didn't list:** `last_completed_task` returns `task_history[-1]` unfiltered, so a failure landing last becomes the lamella's reported latest progress in `experiment_summary_dataframe`. Filtered that too — it now scans backwards for the last *completed* entry.

**Touches your #264.** The four completion-hook tests built history as `AutoLamellaTaskState(name=...)`, which defaults to `NotStarted`, and passed only because nothing filtered. Their docstring anticipated it — *"these break if the predicate's inputs change"* — so the fixtures now say `status=Completed`, which is what they always meant. Worth a look since they're yours.

**Left alone deliberately**, from a sweep of every `task_history` reader:

- `task_history_dataframe` stays unfiltered — showing failures is the entire point of FIB-490.
- `reporting.py:404` / `:661` — a failed task's images and milling patterns now appear in the report. Arguably the more useful behaviour; that's the diagnostically interesting task.
- `autolamella_experiment_task_summary_widget.py:546` labels `len(task_history)` as "completed tasks", so it now over-counts. Real but cosmetic — I'd rather file it than widen a PR that's already changing semantics.
- `manager.py:208` takes timing from the most recent matching entry with status carried separately from the queue. Correct as-is.

patrickcleeve2:
Correcting my framing above, because it invites the wrong follow-up.

I described `is_failure` as a net that *failed to catch* a failed prerequisite. That reads as a gap to close — and closing it would be a worse bug than the one this PR fixes.

`_should_skip` skips a lamella marked failed for **every** remaining task. So auto-setting `is_failure` when a task fails would let one stage timeout or momentary comms drop permanently abandon a lamella that only needed retrying.

A task failing is a fact about one attempt. A defect is a judgement about the lamella. Human-set is correct, and the two should stay separate.

Whether a failed task blocks dependent work is the *other* question, and it's answered where it should be — `completed_tasks` filtering on task status, which is what ea30ee95 does.

Documented on the property itself in aca60376 so the reasoning survives this PR.

PR #282, opened 2026-08-04
[fibsem] Replace the embedded SystemSettings with one hardware geometry record (FIB-481)

patrickcleeve2:
Every image carried a whole `SystemSettings` — **1588 bytes** measured on a real Aquilos file — to deliver six numbers to the reprojection, dragging the manipulator configuration and simulator flags into every picture. v6 records `system_info` + `hardware_geometry` instead: **480 bytes, 70% less**.

## No numerical change

I reconstructed the old formula and ran it against the new code path across model × tilt × rotation × beam, on all four metadata fixtures:

```
compared 560 combinations across 4 fixtures (model x tilt x rotation x beam)
MISMATCHES: 0
```

This moves where the reprojection reads its terms from, and nothing else.

An earlier revision of this PR also fixed a divergence in the compustage FIB-orientation test, which produced 6 sign differences on unreachable poses. **That is now split out to FIB-500**, so a stage-sign change is not buried inside a 20-file restructuring. The site carries a `NOTE` pointing there, and the measurements live on the issue.

## One record for both modalities

`FMImageGeometry` is gone. The beam and FM paths had separate structures naming the same six terms identically, free to drift with nothing to catch it — and they *had* drifted; that is FIB-500. `FibsemHardwareGeometry` now carries the camera's `camera_tilt` and `transform`, left at defaults on a beam image: two dormant fields, 39 bytes, against the 1108 saved.

`CameraImageTransform` and its parser move to `fibsem.structures`, because the record needs the enum and core cannot import from the FM package — `fm.structures` imports from core, so the reverse is a cycle. Both are re-exported from `fm.structures`, beside `AutoFocusMode`, which is re-exported for exactly the same reason.

`fm_image_geometry()` is now `replace()` over `hardware_geometry()`, so the instrument's terms are read out of `SystemSettings` in one place.

## Loading old files

`from_dict` recovers both fields from a pre-v6 `system` blob using `.get()` chains, **never** `SystemSettings.from_dict` — that constructor is bracket-indexed throughout and raises `KeyError` on a blob missing any sub-key, which is exactly what old files risk. Enforced by a parametrised test over all six sub-keys, and guarded against real v3 files from two instruments.

Compustage is recovered by the same model-name match the reprojection used to do inline. That match now survives only inside the shim.

## Also fixed, found while reviewing this

An acquired image **aliased** the live `SystemInfo` rather than copying it, so `TescanMicroscope` rewriting `system.info.model` from each image header reached back into images already acquired. It pre-dates this change (`metadata.system = self.system` aliased the whole object), but the two fields the acquisition helper sets should not disagree about whether they are snapshots. Now both are, with a test.

## Refactor folded in

`_set_additional_metadata` moves to the `FibsemMicroscope` ABC. The same three lines were written out at **eight** acquisition sites across three microscope classes, so changing what an image records meant finding all eight. Happy to split this out if you would rather review it separately.

## Test coverage gap closed

`-128` added to the parity sweep, so the compustage sign flip is exercised at all — `is_fib_orientation` was False in **all 36** combinations before, despite compustage being parametrised. Passes under the existing formula, so this is coverage rather than a behaviour change.

## Known limitation

A v6 server writing metadata to a v5 client leaves `hardware_geometry` unset, and the reprojection raises rather than silently misprojecting. Mixed-version deployments need both sides upgraded.

## Verification

2539 passed, 24 skipped, on top of `3af5a0ae`.

### DIFF
--- a/fibsem/applications/autolamella/ui/AutoLamellaUI.py
+++ b/fibsem/applications/autolamella/ui/AutoLamellaUI.py
@@ -989,7 +989,8 @@ class AutoLamellaUI(QMainWindow):
                     resolution=(2048, 2048), hfw=4000e-6
                 )
                 image.metadata.microscope_state = ms  # type: ignore
-                image.metadata.system = self.microscope.system  # type: ignore
+                image.metadata.system_info = self.microscope.system.info  # type: ignore
+                image.metadata.hardware_geometry = self.microscope.hardware_geometry()  # type: ignore
                 self.minimap_plot_widget.image = image
 
             beam_type = self.minimap_plot_widget.image.metadata.beam_type  # type: ignore
--- a/fibsem/config.py
+++ b/fibsem/config.py
@@ -15,7 +15,11 @@ import fibsem
 # to v5 when it dropped its duplicates of the application/version fields, which
 # SystemInfo owns, and when `application_version` went entirely -- it was null in every
 # file ever written (FIB-448). Those keys are still present in v4-and-earlier files.
-METADATA_VERSION = "v5"
+# v6 replaced the embedded `system` (a whole SystemSettings, 1683 bytes) with
+# `system_info` and `hardware_geometry` (FIB-481). from_dict recovers both from a v5-
+# and-earlier `system` blob, so older files still load -- including their compustage
+# flag, which up to v5 had to be inferred from the model name.
+METADATA_VERSION = "v6"
 # What an unversioned file is. Absent means written before versioning existed, i.e.
 # older than v1 -- not "current", which is what defaulting to METADATA_VERSION claimed.
 UNVERSIONED_METADATA = "v0"
--- a/fibsem/fm/reprojection.py
+++ b/fibsem/fm/reprojection.py
@@ -6,7 +6,7 @@ would send the stage, these ask where the stage already is on screen. Both run t
 projection, so a marker drawn here lands where clicking there would take you.
 
 Microscope-free by design. Everything the projection reads comes from the image's own
-:class:`FMImageGeometry` and stage position, so a saved overview reprojects correctly
+:class:`FibsemHardwareGeometry` and stage position, so a saved overview reprojects correctly
 even when the stage has since moved or the display transform has since been flipped --
 which is precisely when someone is looking at a saved overview. Reading any of it from
 a live microscope would make the result depend on state the image does not describe.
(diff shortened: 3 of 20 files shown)


patrickcleeve2:
**The reachability argument is now a proof rather than evidence.** Pushed as `443c3d27`.

The PR body argued the six divergent poses were unreachable from three observations: no `r` axis in `STAGE_LIMITS_COMPUSTAGE`, the real Arctis image at `r = 0.0`, and 114 corpus records at `r = 0.0`. That is inductive — it says we have never seen one.

The actual guarantee is structural. AutoScript's `CompustagePosition` has no rotation field at all (`x`, `y`, `z`, `a`), so `stage_position_from_autoscript` writes **`r=0.0` as a literal** when converting one:

```python
if isinstance(position, CompustagePosition):
    return FibsemStagePosition(
        x=position.x, y=position.y, z=position.z,
        r=0.0,                       # <- no rotation axis to read
        t=position.a,
        coordinate_system=CoordinateSystem.SPECIMEN.upper(),
    )
```

`FibsemStagePosition` requires a number — `get_stage_orientation` raises on `None` — so zero is how a missing axis gets recorded.

So a compustage image with `r != 0` is not unlikely, it is **unrepresentable**. There is no path from hardware to such a record. The rotation term added to the FIB-orientation test can therefore only change the answer for a pose that cannot be acquired, which is why the six differences do not correspond to any real behaviour change.

Pinned by `test_a_compustage_position_is_converted_with_rotation_hardcoded_to_zero`, asserted against the AST because the function raises without the AutoScript libraries CI does not have. If that literal ever changes, the test fails and this argument gets re-examined rather than silently expiring.

2542 passed, 24 skipped.

patrickcleeve2:
**Rescoped to the boring change.** The compustage FIB-orientation fix is out, split to FIB-500.

Old formula vs new code path, across model × tilt × rotation × beam on all four fixtures:

```
compared 560 combinations across 4 fixtures
MISMATCHES: 0
```

Previously 6, all Arctis at tilt −128 with rotation ≠ 0. Those came from the formula change, not from the restructuring, and they are now FIB-500's problem — with the full measurement, the reachability argument, and the three tests that pin it.

What is left here moves where the reprojection reads its terms from and changes nothing it computes.

Also fixed since the last review: `ast.unparse` is 3.9+ and broke the 3.8 build. That test went to FIB-500 along with the rest, and has been rewritten to inspect AST nodes directly so it works on 3.8.

PR #295, opened 2026-08-05
Correlation: name an image by the file it loaded, not the acquisition request (FIB-509)

patrickcleeve2:
Six places in the correlation UI answer *"which file is this image?"* by reading the acquisition **request** — `image_settings.filename` for FIB, `metadata.filename` for FM. That is the name the acquiring process asked for, and it is not the file.

Closes FIB-509.

## Why the request is the wrong source

**It is a stem.** `FibsemImage.save` appends the suffix itself (`Path(path).with_suffix(".tif")`), so the recorded name is `ref_Mill Fiducial_start` for a file called `ref_Mill Fiducial_start.tif` — confirmed across all four metadata fixtures. `_ImagePicker.select_file` compares basenames:

```python
target = os.path.basename(name)                        # 'ref_Mill Fiducial_start'
if path == name or os.path.basename(path) == target:   # 'ref_Mill Fiducial_start.tif'
```

So the match in `set_fib_image` **has never once succeeded**. The only reason the protocol editor's dialog lands on the right file is that `add_lamella_setup` runs a moment later and re-selects it via `set_image_options(current=...)`.

What that leaves visible is the **header label disagreeing with the picker directly above it** — `set_image_options` does not revisit the label, so it read `ref_Mill Fiducial_start` beside a picker reading `ref_Mill Fiducial_start.tif`. One image, two names, adjacent controls.

**A plain TIFF got no name at all.** `FibsemImage.load` sets `metadata = None` when there is no fibsem `ImageDescription` but sets `filepath` regardless, so an externally-produced file showed a blank label and saved `fib_image_filename: null` — for a file whose path the widget was holding. FIB-316's defensive `getattr` chains stop that crashing the auto-save; they cannot make it say anything.

**A renamed or copied file** went on reporting the name it was acquired under.

## The change

`filepath` is set by `load()` and `save()` on both image classes whatever the metadata says, and answers the filesystem question these sites are actually asking.

| site | now reads |
|---|---|
| `CorrelationInputData.fib_image_filename` / `.fm_image_filename` | `_loaded_filename()` — one helper, `basename(filepath)` |
| `_ImagesTab.set_fib_image` / `.set_fm_image` | `image.filepath` into `select_file` |
| `CorrelationTabWidget._update_fib_name_label` | `filepath`; tooltip is now the full path |
| `FMImageDisplayWidget._update_name_label` | `filepath`; tooltip is now the full path |

**No fallback** to the requested name when `filepath` is unset. An image that was never written has no file, and naming one it does not have sends whoever reads the record looking for it.

**No migration.** `fib_image_filename` / `fm_image_filename` are provenance only — nothing computes on them, and `from_dict` does not restore them (unlike `stored_fib_image_shape` / `stored_fib_image_pixel_size`). Old correlations keep their stems; new ones gain the extension.

## Deliberately not changed

`select_file` returning `""` still leaves the picker alone rather than adding the file via `show_path`. The picker's list means "this lamella's images", and an outside image joining it through a programmatic pre-load is a separate decision from naming. Pinned by a test so it stays deliberate.

## Testing

`tests/correlation/test_image_identity.py` — 11 tests. **10 failed before the fix**; the one that passed is the one pinning why the old match was impossible. `test_the_header_and_the_picker_agree` reproduces the real open sequence (`set_fib_image` then `set_image_options`) and fails on the unfixed source with exactly the stem-vs-extension pair.

- **768 passed, 15 skipped** — `tests/correlation/` + `tests/fm/`
- **2942 passed, 29 skipped** — full suite
- All changed files parse under 3.8 grammar

Two existing tests in `test_correlation_tab_widget.py` pinned the old behaviour and are updated. `_fake_fm` now carries both `filepath` and `metadata.filename`, so the distinction stays visible in the fixture.

## Origin

Split out of FIB-483 and FIB-482, both now cancelled. What they proposed — removing `save`/`path`/`filename` from `ImageSettings` or from the image record — is not wanted: expressing "acquire this, save it there, call it that" as one object is good API, and sharing an image means sharing the path anyway. Migrating the *consumers* to the accurate source is the part that stands on its own.

### DIFF
--- a/fibsem/correlation/structures.py
+++ b/fibsem/correlation/structures.py
@@ -6,6 +6,7 @@ from enum import auto, Enum
 from typing import Optional
 import json
 import logging
+import os
 import time
 
 import numpy as np
@@ -93,6 +94,24 @@ def apply_z_surface_correction(
     return corrected
 
 
+def _loaded_filename(image) -> Optional[str]:
+    """Name the file ``image`` was loaded from, or None if it came from nowhere.
+
+    Read off ``filepath``, not the metadata. The metadata records the name the
+    *acquirer* asked for, which is a stem -- ``save`` appends the suffix itself --
+    and goes stale the moment a file is renamed or copied. ``filepath`` is set by
+    both ``load()`` and ``save()`` on either image class, including for a plain
+    TIFF whose metadata failed to parse at all (FIB-509).
+
+    Deliberately no fallback to the requested name: an image that was never
+    written has no file, and naming one it does not have sends whoever reads the
+    record looking for it. None is the honest answer, and this is read by
+    ``to_dict`` -- the only persistence path -- so it must not raise (FIB-316).
+    """
+    filepath = getattr(image, "filepath", None)
+    return os.path.basename(filepath) if filepath else None
+
+
 @dataclass
 class CorrelationInputData:
     fib_image: Optional[FibsemImage] = None
@@ -173,16 +192,11 @@ class CorrelationInputData:
 
     @property
     def fib_image_filename(self) -> Optional[str]:
-        # Also read by to_dict, so it must tolerate an image without metadata
-        # rather than take the auto-save down with it (FIB-316).
-        settings = getattr(
-            getattr(self.fib_image, "metadata", None), "image_settings", None
-        )
-        return getattr(settings, "filename", None)
+        return _loaded_filename(self.fib_image)
 
     @property
     def fm_image_filename(self) -> Optional[str]:
-        return getattr(getattr(self.fm_image, "metadata", None), "filename", None)
+        return _loaded_filename(self.fm_image)
 
     @staticmethod
     def from_dict(data: dict) -> CorrelationInputData:
--- a/fibsem/correlation/ui/widgets/correlation_tab_widget.py
+++ b/fibsem/correlation/ui/widgets/correlation_tab_widget.py
@@ -628,19 +628,15 @@ class _ImagesTab(QWidget):
 
     def set_fib_image(self, image: FibsemImage) -> None:
         self._fib_image = image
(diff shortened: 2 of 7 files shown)


patrickcleeve2:
Second commit: derived FM volumes were losing the name.

Naming an image from `filepath` means anything that *builds* a new `FluorescenceImage` has to carry it, and three paths did not — `interpolate_fm_volume`, `max_intensity_projection`, `focus_stack`.

Interpolate is a button in the Images tab, so this was reachable. Measured on the first commit:

```
record BEFORE this PR (metadata.filename): fm_stack_as_acquired
record AFTER  this PR (filepath basename): None
```

Interpolating a stack and then correlating wrote `fm_image_filename: null` into the saved correlation and blanked the FM header — worse than what it replaced, on a normal path, in a PR about provenance. CI passed all six jobs on `e84733cf`; nothing existing covered it.

A resampled or projected volume is still that file's volume, and it is the only file that describes it. Same reason those paths already carry `experiment` and `geometry` forward (FIB-466). Safe to carry: nothing writes back through `filepath` — `FluorescenceImage.save` takes the filename it is given, `FibsemImage.save(None)` falls back to the *metadata* path — and the only substantive reader is `base.py:530`, which only sees images straight from `save()`.

`interpolate_fm_volume` reads it via `getattr`, matching how it already treats optional metadata fields: `test_interpolation.py` drives it with a `SimpleNamespace` stand-in carrying only the fields it reads.

Two tests added; both fail without the fix. Full suite **2945 passed, 29 skipped**.

Also checked and clean: `matches_inputs` uses `_transform_inputs`, which excludes image-derived parameters, so correlations saved with the old stem do not start reporting stale against the new name.

PR #296, opened 2026-08-05
[autolamella] Record the session that worked on an experiment — instrument, operator, software (FIB-451)

patrickcleeve2:
An experiment directory records what it contains and nothing about what made it. "Was this run before or after the upgrade?", "which machine was this on?" and "which build of my plugin ran?" are all unanswerable from the record — the one fact all five personas in the project need.

Closes FIB-451.

## The hook

Every part of this was already collected per-image. What was missing was a moment to promote it, and `Experiment.register_metadata(microscope)` is that moment — the only place an experiment meets a microscope, already called by the app when it adopts one, by `TaskManager` when it takes one to run, and by scripts directly (FIB-449). Registration now runs both ways: identity down to the microscope as before, provenance up to the experiment.

A real run records:

```yaml
provenance:
  plugins: {fibsem-plugin-example: 0.1.0}
  recorded_at: 1785907692.139788
  system:
    application: autolamella
    fibsem_revision: v0.5.1-138-g295badb7-dirty
    fibsem_version: 0.5.2.dev0
    manufacturer: Demo
    model: DemoMicroscope
    serial_number: '123456'
  user: {name: Dr Someone, organization: Some Institute, hostname: ...}
```

`installed_plugin_versions()` derives the plugin map from the existing extension report. **Having a distribution is the filter** — built-ins have none and script-loaded extensions carry a path instead. A plugin whose entry point *failed* to load stays in: it was installed, and being absent from the registry shaped the run, so omitting it would make the record match a machine that never had it.

## The last run, not the creating run

`Experiment.create()` has no microscope — the dialog that calls it holds none — so the instrument is genuinely unknown until a run adopts the experiment. Recording the last run is the fact that can actually be established, and it backfills onto experiments that predate this.

Overwriting loses less than it looks like. Every image already carries its own `SystemInfo`, so an experiment spanning an upgrade shows v0.5.1 on day-one images and v0.5.2 here — two true facts, the same reasoning that put the version fields on `SystemInfo` (FIB-445 D1). A per-session history is FIB-452's shape, not this record's.

## Who ran it

A name typed into the create dialog beats `FibsemUser.from_environment()`: on a shared facility login the OS account names the workstation, not the operator, and somebody who typed a name meant it. Only the two fields the dialog offers are overridden — `hostname` stays as the environment reports it, because it answers a different question.

The `SystemInfo` is **deep-copied**. It is live on the microscope — this very call writes `application` to it — and a record that quietly changes after the fact is not a record. Same-object aliasing has bitten this area twice already (FIB-481, FIB-466), so it is pinned by a test.

## The save is conditional, deliberately

First version called `self.save()` unconditionally, on the FIB-490 reasoning that a record nothing persists is not a record. **That broke 69 tests** — not a fixture problem: there are 29 `Experiment(path=...)` construction sites across 19 test modules, 12 of which never create the directory. An `Experiment` without its directory is a legitimate state (FIB-420) that `TaskManager` is routinely driven in, and registration is not the moment to decide it should have one.

So it saves only when `experiment.yaml` already exists. Both `create` and `load` guarantee that, so a real run always writes; an in-memory experiment keeps it in memory for the next real save. Both halves are pinned by tests.

## Testing

`tests/autolamella/test_experiment_session.py` — 16 tests covering what is recorded, the copy-not-reference guarantee, operator-vs-account precedence, last-run-wins, persistence and its deliberate limit, the round trip, an experiment written before this loading as `None`, and the plugin filter (built-ins out, failed-to-load in, missing version recorded as `unknown`).

- **3096 passed, 21 skipped** — full suite
- All changed files parse under 3.8 grammar

## Not done here

The create dialog still writes `user` / `organisation` into the untyped `metadata` dict, and `_declared_user()` reads them from there. Promoting those to typed fields is a UI change and deserves its own issue rather than riding along in this one.

### DIFF
--- a/fibsem/applications/autolamella/structures.py
+++ b/fibsem/applications/autolamella/structures.py
@@ -1052,7 +1052,7 @@ class Experiment:
         experiment.metadata = ddict.get("metadata", {})
 
         # Absent from every experiment written before FIB-451, and from any that no
-        # run has adopted yet. Both mean the same thing -- nothing is known about
+        # session has adopted yet. Both mean the same thing -- nothing is known about
         # what produced this -- and None says that without guessing.
         session = ddict.get("session")
         experiment.session = SessionInfo.from_dict(session) if session else None
--- a/fibsem/applications/autolamella/structures.py
+++ b/fibsem/applications/autolamella/structures.py
@@ -36,10 +36,12 @@ from fibsem.structures import (
     FibsemImage,
     FibsemRectangle,
     FibsemStagePosition,
+    FibsemUser,
     ImageSettings,
     MicroscopeState,
     Point,
     ReferenceImageParameters,
+    RunProvenance,
 )
 from fibsem.utils import configure_logging as _configure_logging
 from fibsem.utils import format_duration
@@ -992,6 +994,7 @@ class Experiment:
     created_at: float = field(default_factory=lambda: datetime.timestamp(datetime.now()))
     task_protocol: 'AutoLamellaTaskProtocol' = field(default_factory=lambda: AutoLamellaTaskProtocol())
     metadata: Dict[str, Any] = field(default_factory=dict)
+    provenance: Optional[RunProvenance] = None
 
     def __init__(self, path: Path,
                  name: str = cfg.EXPERIMENT_NAME,
@@ -1013,6 +1016,10 @@ class Experiment:
 
         self.task_protocol: AutoLamellaTaskProtocol = None # must be set externally
         self.metadata: Dict[str, Any] = metadata if metadata is not None else {}
+        # What produced this: filled in by register_metadata, when a run adopts the
+        # experiment and there is a microscope to ask. None until then -- creating
+        # an experiment happens in a dialog that has no microscope. See FIB-451.
+        self.provenance: Optional[RunProvenance] = None
 
     def to_dict(self, include_protocol: bool = False) -> dict:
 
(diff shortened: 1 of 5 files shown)


patrickcleeve2:
Second commit renames the record: `RunProvenance` → **`SessionInfo`**, and `Experiment.provenance` → `Experiment.session`.

I originally defended "provenance" as the codebase's own vocabulary. That was circular — measured properly, it appears in **zero `.py` files before today**; every use is from this week's metadata work, i.e. mine. And while "run" *is* well established as a noun for one execution (155 uses vs 14 for "session"), it names the execution, not a record about it, and there is no other `Run*` type.

A session is what `setup_session` establishes: a connected instrument and the configuration around it. That is what a reader expects to find inside the name, and it is what is inside — `SystemInfo`, plus who is at the keyboard, plus which plugins are installed. It also pairs with `SystemInfo`, its main constituent.

The key in `experiment.yaml` changes with it, and reads the way the name promises:

```yaml
session:
  plugins: {fibsem-plugin-example: 0.1.0}
  recorded_at: 1785909521.767876
  system:
    application: autolamella
    fibsem_revision: v0.5.1-139-gca97c655-dirty
    manufacturer: Demo
    model: DemoMicroscope
    serial_number: '123456'
  user: {name: Dr Someone, hostname: ..., organization: 'null'}
```

Nothing has shipped with the old name, so there is no migration. **3096 passed, 21 skipped** on the full suite.

One note for FIB-452: this is the *latest* session, singular. If that issue wants a history it should add its own collection rather than widening this field, so a reader of `session` always gets one answer.

PR #302, opened 2026-08-06
[ui] Run the standalone FibsemUI on the quad-view canvas (FIB-406, phase 4a)

patrickcleeve2:
First half of the Microscope-tab cutover ([FIB-406](https://linear.app/fibsemos/issue/FIB-406)). `fibsem_ui` now displays through `MicroscopeViewController` with **no napari viewer**. AutoLamella's Microscope tab is untouched and still runs on napari — that flip is 4b.

## Why split it here

Phases 1–3 were separable because each tab owned its own viewer. That does not hold for the Microscope tab: every widget resolves `self.viewer = parent.viewer`, and `FibsemImageSettingsWidget` *raised* without one, so the widget ports and the shell flip are coupled.

But **two** shells host these widgets, not one — `AutoLamellaUI` and `FibsemUI` (the `fibsem_ui` entry point). Flipping the standalone one first gives the canvas path a live consumer instead of leaving it dead until 4b. That is exactly what made `set_controller()` safe in phase 2: the Lamella Editor exercised the new path from day one.

The cost is that the two widgets carry both paths in the interim. 4b deletes the napari halves.

## How the path is chosen

Per instance, by what the host supplies: `viewer` → napari layers, `view_controller` → canvas. `_view_controller()` returns `None` whenever a viewer is present, keeping the two **strictly exclusive**. `FibsemUI` always builds a controller, so `FibsemUI(viewer=...)` would otherwise run both at once — every image drawn twice, and the alignment area living in the model while `get_alignment_area()` reads it back off an empty layer. There is a test for that shape.

## Also in here

- **Shift+scroll on a beam canvas nudges the working distance** — 1 µm per notch, spinbox updates immediately, the hardware move is debounced so a burst coalesces. Plain scroll still zooms. No acquisition lockout, unlike the FM objective: WD is beam focus, and adjusting it live is *how you focus*.
- **Quad-view selection ↔ SEM/FIB beam radios sync both ways.** Selecting the FM view leaves the radios alone — there is no FM radio, and forcing one would misreport which beam Acquire is about to use.
- **`_teardown_connections()` on both widgets.** The canvases outlive a reconnect; a stale double-click or scroll delivered to a deleted widget raises inside a slot, and PyQt5 turns that into `qFatal` — the process aborts rather than logging ([FIB-329](https://linear.app/fibsemos/issue/FIB-329)). Two tests emit the signals after teardown to prove nothing arrives.

## Worth a reviewer's attention

**The WD spinbox went from 2 to 3 decimals, and that is visible in AutoLamella too.** At 0.01 mm resolution a 1 µm notch rounds away to nothing, so the scroll step forced it. A test pins the two together so they cannot drift apart.

**A removed guard was already dead.** `FibsemImageSettingsWidget` had `if not hasattr(parent, "viewer") and not isinstance(parent.viewer, napari.Viewer)` — `and`, so it needed both to hold, and a parent genuinely missing `viewer` would raise `AttributeError` evaluating the second clause first. It never fired as written.

**napari is still imported**, because the widgets need `napari.layers` for the AutoLamella branch. No `Viewer` is created — checked separately. The import goes in 4b.

**`open_minimap_widget` deliberately keeps its own napari viewer.** The overview surface is being rebuilt on the FM-overview (multi-image) model in phase 3 ([FIB-405](https://linear.app/fibsemos/issue/FIB-405)); porting it here would mean porting it twice.

## Verification

`FibsemUI()` now constructs under `QT_QPA_PLATFORM=offscreen` — impossible before, since napari's GL canvas does not merely warn headless, it aborts the process. So the viewer-less half of the tests runs against a **real** controller and real canvases.

The napari half has to use a stub viewer for that same reason. That is a real limit and it is stated in the test file: those cases pin *which branch runs and what it calls*, not napari's own rendering.

- 25 new headless tests, both paths
- 900 `tests/ui` + 34 canvas widget tests still pass
- 13 mutations tried, 13 caught — including deleting the teardown disconnects, and making `clear_alignment_area` destroy the rect instead of hiding it (which would hand the workflow `None` mid-run, since `update_alignment_area_ui` reads it back straight after the clear)

**Still needs hardware**, on the standalone app: live acquisition on both beams, click and double-click to move, Shift+scroll WD, auto-contrast/auto-focus, detection → mill. [FIB-358](https://linear.app/fibsemos/issue/FIB-358) (FM canvas double-click after the `fm_stable_move` convergence) remains the highest-risk unverified item.

### DIFF
--- a/fibsem/ui/FibsemImageSettingsWidget.py
+++ b/fibsem/ui/FibsemImageSettingsWidget.py
@@ -1,7 +1,6 @@
 import logging
 from typing import Dict, List, Optional, Tuple
 
-import napari
 import numpy as np
 from napari.layers import Image as NapariImageLayer
 from napari.layers import Points as NapariPointLayer
@@ -57,17 +56,19 @@ class FibsemImageSettingsWidget(QtWidgets.QWidget):
         parent: QtWidgets.QWidget,
     ):
         super().__init__(parent=parent)
-        if not hasattr(parent, "viewer") and not isinstance(parent.viewer, napari.Viewer):
-            raise ValueError("Parent must have a 'viewer' attribute of type napari.Viewer")
 
         self.parent = parent
         self.microscope = microscope
-        self.viewer = parent.viewer
-        self.eb_layer: NapariImageLayer = None
-        self.ib_layer: NapariImageLayer = None
+        # Optional viewer. A host that supplies one (AutoLamella today) gets the napari
+        # layer display; a host that supplies a `view_controller` instead (standalone
+        # FibsemUI) gets the quad-view canvas. Exactly one of the two is live per
+        # instance — see `_view_controller` and `uses_napari`.
+        self.viewer = getattr(parent, "viewer", None)
+        self.eb_layer: Optional["NapariImageLayer"] = None
+        self.ib_layer: Optional["NapariImageLayer"] = None
 
         # TODO: migrate to this structure
-        self.imaging_layers: Dict[BeamType, NapariImageLayer] = {}
+        self.imaging_layers: Dict[BeamType, Optional["NapariImageLayer"]] = {}
         self.imaging_layers[BeamType.ELECTRON] = None
         self.imaging_layers[BeamType.ION] = None
 
@@ -76,11 +77,13 @@ class FibsemImageSettingsWidget(QtWidgets.QWidget):
         self.ib_image = FibsemImage.generate_blank_image(resolution=image_settings.resolution, hfw=image_settings.hfw)
 
         # overlay layers
-        self.ruler_layer: Optional[NapariPointLayer] = None
-        self.alignment_layer: Optional[NapariShapesLayer] = None
+        self.ruler_layer: Optional["NapariPointLayer"] = None
+        self.alignment_layer: Optional["NapariShapesLayer"] = None
 
         self.is_acquiring: bool = False
         self._ruler_enabled: bool = False
+        self._live_beam: Optional[BeamType] = None  # beam currently live (canvas LIVE badge)
+        self._overlay_edited_wired = False  # subscribed to controller.overlay_edited
 
         self._setup_ui()
         self.setup_connections()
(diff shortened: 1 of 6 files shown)


patrickcleeve2:
Pushed `1b9a5c71` — three defects found by running the standalone app, none of which the original tests caught. Each now has one that bites (verified by mutation).

**1. Restore the truncated `setup_connections`** — mine, this branch. I inserted `_teardown_connections` into the *middle* of `setup_connections`, so everything from the instructions label down to the closing `update_ui()` silently became teardown code: stage limits, saved positions, the movement/acquisition signal wiring, milling-angle controls, wheel blockers. The widget still constructed and looked right, which is exactly why 25 new tests and 900 existing ones all passed. The reported symptom was no stage position on the canvas — `update_ui()` is the only thing that puts one there before the operator moves or acquires. Teardown would also have run all that *setup* on a dying widget, including `update_ui()`, which is the qFatal path it exists to prevent.

**2. `MillingTaskViewerWidget.is_milling`** — pre-existing, affects both paths. `FibsemMovementWidget` blocks click-to-move on `parent.milling_widget.is_milling`; that attribute lives on the embedded `FibsemMillingWidget2`, not on the task viewer the hosts actually store, so it raised `AttributeError` inside the worker and every double-click died there. `AutoLamellaUI` has no `milling_widget` attribute at all, so `hasattr` short-circuits and it never fired there — only `FibsemUI` sets one. **The identical line is in the napari `_double_click_worker`, so click-to-move has been broken in the standalone app on `main` too**; the canvas path only made it visible because `FunctionWorker` logs where napari's `thread_worker` reraised. Fixed by delegating from the widget the hosts hold, which repairs both paths.

**3. Gate click-to-move on the widget's own enabled state** — pre-existing, both paths. `_toggle_interactions` disables the move buttons for the whole move *and* the acquisition that follows it (`move_stage_finished` deliberately returns early while `is_acquiring`). Canvas input was never gated, so a double-click in that window started a second overlapping stage move and a second acquisition on the same microscope.

That last one is how a SEM frame ended up on the FIB canvas. Worth being precise about what this does and does not fix: tracing showed the display routes correctly every time — the images handed to it were already mislabelled relative to their pixels. Two concurrent `acquire.acquire_image` calls are clean 12/12; it takes the widget's `take_reference_images` path to reproduce. That underlying defect is [FIB-514](https://linear.app/fibsemos/issue/FIB-514) and is **not** fixed here — this only closes the double-click route into it. The Acquire buttons, workflow post-move acquisitions and live imaging all still drive the same widget.

32 tests in the new file, 900 `tests/ui`, 34 canvas widget tests. One fidelity fix along the way: my test hosts weren't setting `movement_widget`, which the image widget reaches back through when toggling interactions — so they'd have passed things the app fails. Routed all the movement tests through a helper that sets it.

PR #305, opened 2026-08-06
[ui] Follow the microscope across a reconnect in the protocol editors (FIB-525)

patrickcleeve2:
Fixes FIB-525.

## The bug

The two protocol editors are built once and survive a disconnect — unlike the main tabs, which `update_microscope_ui` destroys and rebuilds on every connect. So every widget under them keeps whatever microscope it was constructed with. On reconnect, the handlers updated exactly one attribute:

```python
self.milling_task_editor.microscope = self.microscope
```

Walking the widget tree finds **twelve references** to the microscope (or its `fm`) across eleven widget classes, identical in both editors. Two were refreshed. Ten were not.

**Reported symptom**: connect to a system with no fluorescence detector → disconnect → reconnect to an FM-enabled one, and the channel editor's excitation/emission dropdowns stay empty for the life of the process. The wavelengths are read from the microscope once, at construction, so an FM-less first connection caches two empty lists.

**The mirror case is worse**, and is the better reason to fix it: connect to an FM microscope *first*, then reconnect to a different one, and the widget keeps a live handle to the **previous microscope's `fm`**. Same for `FibsemMillingWidget2`, which doesn't merely read the microscope — it drives milling and subscribes to its progress signal.

## Also: the existing one-liner was never sufficient

`MillingTaskViewerWidget` hands the microscope to two children at construction, and the available currents are constructor arguments one level below *that*. So the milling side had the same fault, masked because currents and presets only visibly differ when the new microscope is a different make.

| chain | widget | refreshed before |
|---|---|---|
| milling | `<Editor>.microscope` | yes |
| | `MillingTaskViewerWidget.microscope` | yes |
| | `MillingTaskConfigWidget2.microscope` | **no** |
| | `FibsemMillingStagesWidget.microscope` | **no** |
| | `FibsemMillingSettingsWidget.microscope` | **no** |
| | `FibsemPatternSettingsWidget.microscope` | **no** |
| | `FibsemMillingWidget2.microscope` | **no** |
| fluorescence | `…FluorescenceAcquisitionTaskConfigWidget.microscope` / `.fm` | **no** |
| | `FluorescenceMultiChannelWidget.fm` | **no** |
| | `ChannelListWidget.fm` | **no** |
| | `_DraggableChannelList._fm` | **no** |

## The fix

Each widget holding a microscope gains `set_microscope` (or `set_fm`), re-deriving what it cached and passing it down:

- **Fluorescence**: re-reads the wavelength lists and rebuilds the channel rows, which take their choices as constructor arguments. The channel objects are untouched, so the configured channels survive; row selection and the enabled checkboxes don't — an acceptable cost for swapping microscopes.
- **Milling**: re-reads available currents and presets, and `show_preset` — whether presets apply at all is a property of the connected system (Tescan mills by preset), not a fixed layout choice.
- **`FibsemMillingWidget2`**: moves its progress subscription to the new microscope instead of leaving it on the old object.

Where a widget already had a clear-and-rebuild path (`FibsemPatternSettingsWidget._build_controls`), the setter reuses it rather than inventing a second one — the pattern and strategy forms both document a segfault from removing form rows during a focus change, and this doesn't add another place that can hit it.

## Tests — `tests/ui/test_editor_microscope_reconnect.py`

The primary test walks the tree after a simulated reconnect and asserts **nothing still points at the old microscope**. Deliberately not a list of widget names: an incomplete list is precisely what produced this bug, and a future widget that captures a microscope fails here without anyone remembering to add it. It also asserts the tree *does* hold the new one, so it can't pass by everything being dropped.

That alone isn't sufficient, though. Two widgets keep only values **derived** from the microscope and no reference at all — `ChannelSettingsWidget`'s wavelength lists, `MillingStageListWidget`'s currents — so an identity walk is blind to them. Those are asserted against what the new microscope reports.

8 tests, both editors. Mutation tested — each caught:

| mutation | caught by |
|---|---|
| fluorescence widget not told about the new microscope | tree walk + channel filters |
| milling subtree reverted to the bare assignment (the old behaviour) | tree walk + milling currents |
| channel list keeps its cached wavelengths | channel filters ×2 |
| detail widget keeps its cached wavelengths | channel filters ×2 |

`tests/ui` + `tests/milling` + `tests/autolamella`: **1513 passed**. All changed files parse under 3.8 (CI matrix floor).

## Not addressed

The editors don't listen to `disconnected_signal` at all, so between a disconnect and the next connect they still hold a handle to a disconnected microscope. Out of scope here — this PR makes reconnect correct; deciding what the editors should *do* while disconnected (disable? clear?) is a separate question.

### DIFF
--- a/fibsem/applications/autolamella/ui/autolamella_fluorescence_acquisition_task_config_widget.py
+++ b/fibsem/applications/autolamella/ui/autolamella_fluorescence_acquisition_task_config_widget.py
@@ -84,6 +84,22 @@ class AutoLamellaFluorescenceAcquisitionTaskConfigWidget(QWidget):
 
         self.setLayout(layout)
 
+    def set_microscope(self, microscope: 'FibsemMicroscope') -> None:
+        """Point this widget at a different microscope after a reconnect.
+
+        The editors that own this widget are built once and survive
+        disconnect/reconnect, so without this the widget keeps the microscope it
+        was constructed with -- and its channel editor keeps offering that
+        microscope's excitation/emission filters. Connect to a non-FM system
+        first and it offers none at all, for the life of the process (FIB-525).
+
+        The captured ``fm`` matters as much as the microscope: it is a live
+        handle to hardware, not a value.
+        """
+        self.microscope = microscope
+        self.fm = microscope.fm
+        self.channelSettingsWidget.set_fm(microscope.fm)
+
     def get_task_config(self) -> AcquireFluorescenceImageConfig:
         self.config.orientation = self.orientation_combo.value()
         self.config.zparams = self.z_parameters_widget.z_parameters
--- a/fibsem/ui/fm/widgets/channel_list_widget.py
+++ b/fibsem/ui/fm/widgets/channel_list_widget.py
@@ -95,6 +95,15 @@ class _DraggableChannelList(QListWidget):
         super().__init__(parent)
         self._fm = fm
 
+    def set_fm(self, fm) -> None:
+        """Point at a different microscope (FIB-525).
+
+        Only used to suppress selection changes during live acquisition, but a
+        stale reference here means asking the *previous* microscope whether it is
+        streaming.
+        """
+        self._fm = fm
+
     def resizeEvent(self, event) -> None:
         """Keep row widths pinned to the viewport as it changes.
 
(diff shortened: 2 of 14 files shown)


patrickcleeve2:
Superseded by #306, which is the fluorescence half only.

Splitting because this PR bundled two things: the reported bug (fluorescence channel editor, small and self-contained) and the same fault in the milling subtree, which touches the widget that drives milling and deserves its own review. The milling half is now FIB-530, with the measured list of stale references and the two traps found here — the displaced `@ensure_main_thread` decorator, and the interpreter abort from emitting change signals during a refresh.

The branch is left in place as a starting point for FIB-530.

patrickcleeve2:
Correction: the replacement is **#307**, not #306.

PR #309, opened 2026-08-07
[ui] Accept one metadata vocabulary across all four config forms (FIB-384)

patrickcleeve2:
Fixes FIB-384. First step of the **Config Forms and Plugin Widgets** project.

## The problem, measured

`DEFAULT_FIELD_METADATA` is the vocabulary. Counting what each form actually reads:

| form | keys read |
|---|---|
| pattern | 15 |
| strategy | 15 |
| milling settings | 14 |
| **task parameters** | **4** — `items`, `scale`, plus its own `units` and `help` |

So it isn't a three-way divergence as the issue framed it — the milling family already agrees, and the task form is the outlier: it reads a quarter of the vocabulary, and the two keys it does read for text and units are spelled differently from everywhere else.

A field written for one form therefore renders half-blank in the other — right value, no tooltip, no unit suffix, nothing logged. The clearest in-tree proof is `microscopes/_stage.py`, which declared **both dialects in one metadata dict**:

```python
metadata={"units": "mm", "tooltip": "Radius of the sample grid", "scale": 1e3}
```

Whichever form rendered that, one of those two was always ignored.

## The change: one vocabulary, not two accepted forever

- **All 37 in-tree declarations renamed** to `tooltip` / `unit`, across 11 files.
- **The task form reads the canonical keys directly.** No alias resolution anywhere — `get_fields_with_metadata` resolves nothing, and the old spellings are simply not read.
- **A key nothing reads is now logged once** per (class, field, key). Once, because form metadata is re-read on every rebuild — a repeating warning was the exact noise problem fixed in #301.
- **A *superseded* key gets a targeted warning**: a field still declaring `help` is told it was renamed to `tooltip` and that the field will render without it, rather than getting the generic "no form reads this" line. `RENAMED_METADATA_KEYS` exists only to phrase that message; nothing resolves it.
- **`filepath` is now in the vocabulary.** Both milling forms read it; it was never listed, which made it undiscoverable to exactly the plugin authors this exists for.

An earlier revision of this PR kept `help`/`units` as permanent aliases. That traded a one-time conversion for a compat surface carried indefinitely, and the argument for it was weaker than it looked — **the warning is what makes dropping them safe rather than silent**. A plugin that hasn't converted loses a tooltip and is told precisely why.

## Verification

- **The rename is behaviour-preserving**: resolved metadata for all 19 registered config classes (90 fields) snapshotted before and after, asserted equal.
- **No rendering change for existing configs**: label tooltips and control suffixes for every registered task config byte-identical to `main`.
- **Adding `filepath` changed no control**: every registered pattern and strategy rendered, all 149 rows diffed against `main` — identical, with the single `QFilePathLineEdit` still the only one. Its default must stay falsy, since both forms select the file picker with a truthy `m.get("filepath")`. Pinned by a test.
- The in-tree sweep iterates `BUILTIN_TASKS`, not `get_tasks()` — the latter includes whatever plugins are pip-installed, so it passed or failed on the contents of a developer's environment.

Full suite **3348 passed**. Decorator audit clean. All changed files parse under 3.8.

## Companion change

`fibsem-os/fibsem-plugin-example` is updated in the same breath (committed locally, not yet pushed): it declared the old spellings and, worse, carried a comment documenting the divergence as a fact of life. It's the pattern plugin authors copy, so it shouldn't demonstrate keys that stopped working.

The other known out-of-tree plugin (a lab CLEM plugin) still declares `help`/`units`; it will lose those tooltips and log the targeted warning until converted.

## Follow-ups

- **FIB-531** — declare metadata through a typed constructor, so a wrong key is a `TypeError` at import rather than a warning at runtime. Also converts the example repo to that API.
- **FIB-526** — the larger win, untouched here: the task form still reads only four keys, so `minimum`/`maximum`/`step`/`decimals`/`hidden`/`advanced` remain unavailable to task configs.

### DIFF
--- a/fibsem/applications/autolamella/workflows/tasks/acquire_fluorescence.py
+++ b/fibsem/applications/autolamella/workflows/tasks/acquire_fluorescence.py
@@ -27,19 +27,19 @@ class AcquireFluorescenceImageConfig(AutoLamellaTaskConfig):
     task_type: ClassVar[str] = "ACQUIRE_FLUORESCENCE_IMAGE"
     display_name: ClassVar[str] = "Acquire Fluorescence Image"
     channel_settings: list[ChannelSettings] = field(default_factory=list,
-                                                    metadata={"help": "Settings for each fluorescence channel",
+                                                    metadata={"tooltip": "Settings for each fluorescence channel",
                                                              "label": "Channel Settings"})
     zparams: ZParameters = field(default_factory=ZParameters, 
-                                  metadata={"help": "Z-stack acquisition parameters",
+                                  metadata={"tooltip": "Z-stack acquisition parameters",
                                             "label": "Z-Stack Parameters"})
     autofocus_settings: AutoFocusSettings = field(default_factory=AutoFocusSettings,
-                                                  metadata={"help": "Settings for autofocus before acquiring fluorescence images",
+                                                  metadata={"tooltip": "Settings for autofocus before acquiring fluorescence images",
                                                             "label": "Autofocus Settings"})
     orientation: Optional[str] = field(default=None,
-                                       metadata={"help": "Orientation for acquisition. 'FM' or 'SEM'. None = use fluorescence_pose as-is.",
+                                       metadata={"tooltip": "Orientation for acquisition. 'FM' or 'SEM'. None = use fluorescence_pose as-is.",
                                                  "label": "Orientation"})
     retract_objective: bool = field(default=True,
-                                    metadata={"help": "Retract the objective when the task finishes, so it is clear of the stage for subsequent moves. Disable for back-to-back fluorescence tasks.",
+                                    metadata={"tooltip": "Retract the objective when the task finishes, so it is clear of the stage for subsequent moves. Disable for back-to-back fluorescence tasks.",
                                               "label": "Retract Objective"})
 
     def to_dict(self) -> dict:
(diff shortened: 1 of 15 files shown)


patrickcleeve2:
Review follow-ups pushed as `bd2ca5d1`.

- **`filepath` is now genuinely in `DEFAULT_FIELD_METADATA`.** The body claimed it was; the code had it in a private `_EXTRA_KNOWN_METADATA_KEYS` set that only suppressed the warning, leaving it undiscoverable to plugin authors. That private set is gone.

  The default has to stay falsy — both milling forms select the file-picker control with a truthy `m.get("filepath")`, so a truthy default would turn every string field into a file picker. Verified by rendering every registered pattern and strategy and diffing the control type of all **149 rows** against `main`: identical, with the single `QFilePathLineEdit` still the only one. Pinned by a test.
- **`metadata_value` is typed `Mapping[str, Any]`**, not `Dict` — every caller passes `dataclasses.Field.metadata`, which is a `mappingproxy`.
- Dropped an unused `ClassVar` import from the new test module.

Full suite: **3348 passed**.

Still open from the review, deliberately not in this PR:
- the unknown-key warning assumes fibsem's forms are the only readers, so a plugin shipping its own config widget could get spurious warnings for keys its own widget reads. Worth settling alongside FIB-528.
- `known` is rebuilt per field inside `_warn_unknown_metadata_keys`; trivial cost, wants hoisting.
- the rename's correctness rests on a one-time before/after snapshot (19 config classes, 90 fields), not on a durable test — the suite would catch a nonsense key but not a valid-but-wrong one.

PR #328, opened 2026-08-07
[ui] Fix task parameter edits being silently dropped (FIB-526 regression)

patrickcleeve2:
**Regression from #320.** Editing a task parameter in the protocol editor did nothing — the value reverted as soon as you switched tasks and came back, because it never reached the config at all.

## Root cause

#320 captured the field name for the change handler like this:

```python
lambda name=row.field: self._on_parameter_changed(name)
```

That lambda accepts **one positional argument**. PyQt sizes the call to whatever the slot will take, so it passed the signal's payload straight into `name`:

| signal | payload the handler received instead of the field name |
|---|---|
| `valueChanged` | `42.0` |
| `toggled` | `False` |
| `currentIndexChanged` | `1` |
| `editingFinished` | *(none — so text fields kept working)* |

The handler then looked for a row whose field was `42.0`, found none, and returned. No error, no log line, nothing written, nothing emitted — so `_on_task_parameters_config_changed` in the protocol editor never fired and `_save_experiment()` never ran.

That last row is why it looked like it half worked: only `editingFinished` carries no payload.

The previous implementation used a zero-argument `lambda:` and was immune. I introduced the default argument to capture the loop variable and didn't notice it changed the slot's arity.

## The fix

In `Control.connect`, not at the two call sites: the slot is now always invoked with no arguments.

```python
signal.connect(lambda *_: slot())
```

A form never wants these payloads — it calls `read()` — and the signals disagree about what they carry, so normalising once is what stops the next caller hitting this. A slot that genuinely cannot take zero arguments now raises instead of being silently misbound.

## Why the tests missed it

They called `form._on_parameter_changed(name)` **directly**, which exercises the write but not the wiring — and the wiring is what broke. Every assertion about the write-back passed while the form was completely disconnected.

Added, all driving the real signals:

- one per control type, edit → assert the config changed
- the `parameter_changed` payload the protocol editor depends on
- the reported symptom end to end: edit, switch task, switch back, assert the value survived

Reverting either half of the fix fails **11** of them; reverting the `Control.connect` guarantee alone fails 12.

## Scope

The pattern, strategy and milling-settings forms were **never affected** — they connect a bound method taking no arguments. Verified their write-back still works. 1,654 tests pass.

### DIFF
--- a/fibsem/ui/widgets/autolamella_task_config_widget.py
+++ b/fibsem/ui/widgets/autolamella_task_config_widget.py
@@ -188,7 +188,9 @@ class AutoLamellaTaskConfigWidget(QWidget):
                 # A read-only row displays a value it cannot reconstruct; wiring
                 # it up would let a focus-out write the displayed text back.
                 if row.control.editable:
-                    row.control.connect(lambda name=row.field: self._on_parameter_changed(name))
+                    row.control.connect(
+                    lambda name=row.field: self._on_parameter_changed(name)
+                )
             self._update_visibility()
             self.task_params_collapsible.show()
         else:
@@ -304,7 +306,9 @@ class AutoLamellaTaskParametersConfigWidget(QWidget):
             # A read-only row displays a value it cannot reconstruct; wiring it
             # up would let a focus-out write the displayed text back.
             if row.control.editable:
-                row.control.connect(lambda name=row.field: self._on_parameter_changed(name))
+                row.control.connect(
+                    lambda name=row.field: self._on_parameter_changed(name)
+                )
         self._update_visibility()
         self.show()
 
--- a/fibsem/ui/widgets/form_builder.py
+++ b/fibsem/ui/widgets/form_builder.py
@@ -101,9 +101,23 @@ class Control:
             self.inputs = (self.widget,)
 
     def connect(self, slot: Callable[[], None]) -> None:
-        """Wire every signal that means "the user edited this"."""
+        """Wire every signal that means "the user edited this".
+
+        The slot is always called with **no arguments**. The signals behind a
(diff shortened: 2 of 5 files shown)


patrickcleeve2:
**Audit for leakage into the other widgets — it didn't, and there is now a guard.**

The silent form of this bug needs a slot with default-valued parameters and no `*args`, so I swept every `.connect(...)` in `fibsem/` by AST for that shape. **112 sites take a lambda with parameters; 4 use the capture idiom:**

| site | verdict |
|---|---|
| `autolamella_task_config_widget.py:191, :309` | the two this PR fixes |
| `workflow_timeline_widget.py:874` | **correct** — `lambda _checked=False, a=action_id:` |
| `tests/test-selected-lamella.py:82` | **correct** — `lambda _=False, lam=lamella:` |

Both of the others put a sacrificial first parameter to absorb the payload and keep the captured value in second position. That is the form I should have used. The remaining 108 either take the payload deliberately or would raise loudly rather than misbind.

The pattern, strategy and milling forms connect a bound method taking no arguments, so they were never exposed. Verified at runtime anyway — spinbox, combo, checkbox and compound-Point edits all reach their config in every form.

**The runtime check isn't a guard, so `tests/ui/test_form_writeback.py` now is.** Nine tests that drive each form's real signals and assert the config changed, rather than calling the handler. Mutation-checked:

| mutation | result |
|---|---|
| unwire the signals | 9 fail |
| float adapter forgets the scale | 1 fails |
| int adapter returns a float | 1 fails |
| compound row drops the y signal | 1 fails |

That last one is worth flagging: my first draft only drove the Point row's x spinbox and **passed** with y unwired. Both axes are now covered.

1,663 tests pass.

PR #341, opened 2026-08-10
[refactor] Delete two unreferenced widgets

patrickcleeve2:
Found by a dead-code sweep of the repo. **−1242 lines, no behaviour change.**

| LOC | file | |
| ---: | --- | --- |
| 664 | `ui/widgets/autolamella_experiment_task_summary_widget.py` | `ExperimentTaskSummaryWidget`, `create_experiment_task_summary_widget` |
| 578 | `ui/widgets/image_annotation_widget.py` | `ImageAnnotationWidget`, `create_image_annotation_dialog`, `plot_image_with_scalebar` |

Neither is constructed anywhere, neither is exported from any `__init__.py`, and neither has ever been reachable from the app. Every symbol in both files has zero references outside its own module.

## Knock-on

`plot_experiment_task_summary` and `plot_lamella_task_workflow_summary` in `applications/autolamella/tools/reporting.py` had exactly one consumer between them — the summary widget. **They are unreferenced as of this PR.**

Left in place deliberately: `reporting.py` is the natural home for figure generation and the experiment-record work (FIB-445..462) may well want them. But they are dead until something calls them, and that should be a decision rather than a drift.

## Also worth noting

The summary widget sat in the wrong tree — `fibsem/ui/widgets/` while importing `fibsem.applications.autolamella.structures`, so a generic UI module depended on an application. Deleting it removes one instance of that inversion; it is not the only one.

## Verification

- No residual references to any of the five symbols or either module name, in any file type.
- `fibsem.ui`, `fibsem.ui.widgets`, `AutoLamellaMainUI` and `tools.reporting` all still import.
- Full suite: **3589 passed, 34 skipped.**

One deselect: `tests/test_no_acquisition_time_gamma.py::test_no_shipped_configuration_still_declares_it` fails identically on clean `main` in this checkout. It is **not related to this PR** — the test globs `fibsem/config/*.yaml` on disk, which picks up gitignored, runtime-written local configs (here a `coincidence-milling-config.yaml` from Aug 3). CI never has those files, which is why CI is green. Worth fixing separately: a test that claims to check "shipped configuration" should read the shipped set, not the working directory.


patrickcleeve2:
Correction to the "Knock-on" section above — it overstated the effect.

I wrote that `plot_experiment_task_summary` **and** `plot_lamella_task_workflow_summary` were both left unreferenced. Only one of them is:

| function | on `main` after this merge |
| --- | --- |
| `plot_experiment_task_summary` (133 lines, `:525`) | **dead** — definition only, no caller anywhere |
| `plot_lamella_task_workflow_summary` (`:392`) | **still live** — called at `reporting.py:238` |

`reporting.py` itself is healthy: two public entry points (`generate_report2`, `save_final_overview_image`) with three importers, and everything else reachable from them. `plot_experiment_task_summary` is the single orphan.

Still left in place deliberately, for the same reason as before — the experiment-record work is the natural consumer. But it is one function, not two.

PR #394, opened 2026-08-13
[ui] Draw the overview canvas before the first image (FIB-616)

patrickcleeve2:
Opening the Overview tab showed a black rectangle. No travel limits, no grid boundary, no holder slots, no lamella markers, not even the stage — everything appeared at once when the first tile landed, which is the moment it stopped being the most useful. The fluorescence overview has never behaved this way.

Stacked on #388.

## Why

A frame needs three things, and two of them only arrived with an image:

| ingredient | set by | needed an image? |
| -- | -- | -- |
| the view's origin | `_set_origin_from`, called only from `place_image` | **yes** |
| `canvas.reference_pixel_size` | `place_image` | **yes** |
| the projection | `BeamStageProjection.from_microscope` | no |

No frame meant `_refresh_context_overlays` returned early, and every marker with it.

Both are free to supply. The scale is `hfw / width` off the settings widget — a widget read, not a device one — and the origin is the stage position already cached for the markers. `FMOverviewWidget` does the equivalent (`self._origin or self._current_stage_position()`, plus seeding the scale from the camera).

## Three things that had to be right

**The anchor is fixed, not re-derived per call.** Otherwise the stage marker would be pinned to the middle and the travel envelope would slide past it — the wrong way round for overlays that describe where the stage can go.

**It is provisional.** The first real image in the view replaces it, so a canvas that ends up holding data is anchored on the data rather than on wherever the stage happened to be when the tab was opened, which can be millimetres away. Nothing is placed against it at that point, so re-anchoring moves nothing.

**The displayed view follows the stage only until something decides otherwise** — an image arriving or the user picking one — tracked as an explicit pin. Inferring it from "the canvas is empty" does not work: `show_view` clears the canvas before re-placing and refreshes the overlays while it is empty, so the seeding pulled the view back to the stage's half way through switching to another one. That cost three existing tests, which is how it was found.

## Which pixel size wins

It does not matter beyond the units. The reference is only "how many metres a canvas pixel is worth"; `add_image` scales every image by its own pixel size against it, so an image coarser or finer than the plan is still drawn covering the ground it images. There is a test for exactly that, with an image whose field of view is deliberately unlike the settings'.

## Structure

Seeding lives in `_refresh_context_overlays` — the one place that runs on construction, on a stage move, on a settings change and on a view change, which is every moment the answer could have changed. `_frame` stays a pure read, so the many callers that only want to draw or resolve a point cannot quietly re-anchor the canvas by asking.

## Testing

7 new tests; 1323 pass across `fibsem/applications/autolamella/ui/tests/` and `tests/ui/`.

Three mutations, each caught by the test that should catch it:

| mutation | killed by |
| -- | -- |
| no seeding at all | 6 of the 7 |
| provisional anchor never replaced | `test_the_first_image_replaces_the_stand_in_anchor` |
| an image does not pin the view | `test_the_view_follows_the_stage_until_an_image_pins_it` |

`test_each_view_keeps_its_own_origin` now looks its two views up by name rather than counting `_origins` — the provisional anchor for the view the stage is in is a third entry, and it is meant to be there.

### DIFF
--- a/fibsem/ui/widgets/canvas/real_space_canvas.py
+++ b/fibsem/ui/widgets/canvas/real_space_canvas.py
@@ -122,12 +122,7 @@ class FibsemRealSpaceCanvas(FibsemCanvasBase):
         if reference_pixel_size:
             self._pixel_size = reference_pixel_size  # drives the scalebar
 
-        # Square pixels from the outset, not just once an image happens to arrive.
-        # imshow sets this per-artist, so an empty canvas would otherwise be "auto":
-        # the axes would stretch to the widget and draw a square grid as a rectangle --
-        # a resize alone distorted it 3x. Overlays are drawn in this frame whether or
-        # not anything has been acquired, so the frame has to be honest first.
-        self._ax.set_aspect("equal", adjustable="box")
+        self._apply_axes_convention()
 
         self.set_background_color(_DEFAULT_BACKGROUND)
         # Contrast/gamma acts on a single frame; there is no meaning yet for "the"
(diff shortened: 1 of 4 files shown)


patrickcleeve2:
Added the pin release, per discussion.

Removing the **last** overview clears `_view_pinned`, so the canvas goes back to following the stage — an empty canvas pinned to a view nothing is in is the state this seeding exists to avoid, and the image that pinned it is gone. Cleared before the overlay refresh, so it re-points in the same pass.

Two boundaries that needed pinning down:

* **Removing one of several does not release it.** The user is still looking at data; handing the view back to the stage underneath them would be worse than the original problem.
* **Hiding is not removing.** `set_overview_visible` leaves the record, and a view you can bring back with a checkbox is still one you chose to be in.

Worth noting on the second bullet: the `if not self._records` guard was initially untested — clearing the pin unconditionally passed the whole suite. `test_removing_one_of_several_keeps_the_view` is what makes it mean something, and it fails against that mutation.

1326 pass locally.

patrickcleeve2:
Good catch, and it was not a label-placement quirk — the labels were the symptom.

**With nothing placed, the whole scene was drawn vertically mirrored**, and flipped the right way up the moment the first tile landed. A lamella at +100 µm appeared where −100 µm belongs.

```
EMPTY    ylim: (-7509.3, +7509.3)   inverted: False
WITH IMG ylim: (   87.4,    -87.4)  inverted: True
```

`imshow` sets the y direction per artist, so a canvas holding images was right *by accident* and an empty one inherited matplotlib's default of y increasing upwards. Nothing showed it while an empty canvas drew nothing — it appeared as soon as the overlays did, which is this PR.

This is the same shape as the aspect ratio, which is also set per artist and was already forced from the outset for the same reason. Both now live in one `_apply_axes_convention`, called at construction and again after `clear` — `cla()` wipes both.

Two things worth flagging for review:

**The extents were never wrong.** `test_positive_x_is_right_and_positive_y_is_down` asserts them and passed throughout, because it checks the numbers rather than where they land. The new tests go through `transData` instead. `test_y_still_runs_down_once_an_image_is_placed` deliberately still passes against the mutation — that asymmetry *is* the bug being documented.

**My render harness hid it.** It set the axis limits by hand, and setting them descending inverted the axis as a side effect — so the picture I checked looked right while the real tab did not. The harness calls `reset_view` now, like the app.

1329 pass locally.

PR #403, opened 2026-08-13
Treat an unreadable refractive-index lookup table as missing (FIB-592)

patrickcleeve2:
Fixes FIB-592.

A truncated download, or a proxy error page saved under the CSV's name, leaves a file on disk. `lut_available = _LUT_PATH.exists()` was true for it, so the RI panel enabled all five optical-parameter spinboxes, rendered no warning, and every lookup raised into a bare `except Exception` that did not even log. Turning the knobs changed nothing — permanently and silently, because the file exists so nothing ever re-fetched it.

Measured on the same widget with the same edits:

| LUT state | NA 0.8→0.5, n2 1.4→1.22 | panel |
| --- | --- | --- |
| healthy | factor 1.595 → 1.260 | controls enabled |
| truncated | 1.470 → 1.470 | enabled, no warning, no log line |
| HTML 403 page | 1.470 → 1.470 | enabled, no warning, no log line |

This matters more since #373: placing an FM surface point now arms whatever the factor box holds, so a silently-dead calculator means the untouched default reaches a real correlation run.

## Changes

- **`lut_error()` replaces the existence test.** It parses, caches the verdict, and returns *why* the LUT is unusable. The widget reports that reason in the panel and disables the controls, exactly as it already did for an absent file.
- **`_download_lut` writes to a temp file beside the destination and `os.replace()`s it in**, so an interrupted transfer cannot leave a truncated CSV at the real path to begin with.
- **`_ensure_lut` re-fetches a file that will not parse**, so an install that already picked up a bad one self-heals instead of needing it deleted by hand.
- **`_recompute` logs the swallowed exception.**
- **`SliceScalingFactorLUT` resolves `_LUT_PATH` at call time.** As a default argument it bound at import, so the class read the original file after anything repointed the module — a silent disagreement with `lut_error()` rather than a failure. It also now names a missing column instead of dying on a `KeyError`.

## Tests

14 tests against a synthetic LUT in `tmp_path`, in their own module. The existing LUT test skips wholesale when the real CSV is absent, which is precisely the situation these need to cover.

Note for anyone touching this area: nine test files under `tests/correlation/` monkeypatch `riw._ensure_lut` to block the network. Rename or bypass that seam and all nine silently stop protecting, and the suite starts downloading.

## Two open questions, neither blocking this fix

Both are about the download existing at all, not about the bug:

1. **Ship the ~9 MB lookup table as package data and drop the runtime download.** That removes this failure mode entirely rather than handling it, along with the network dependency and the startup latency.
2. **Make the download opt-in.** It still fires automatically when the correlation window opens — on the GUI thread, now bounded by a timeout and attempted once per process. Microscope PCs are commonly offline or behind a proxy that blackholes rather than refuses.

Happy to follow up with either; this PR is the defensive fix on its own.

### DIFF
--- a/fibsem/correlation/refractive_index.py
+++ b/fibsem/correlation/refractive_index.py
@@ -26,6 +26,9 @@ Usage::
 from __future__ import annotations
 
 import logging
+import os
+import shutil
+import tempfile
 import urllib.request
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


patrickcleeve2:
Closing unmerged for now — parking this until the underlying approach is settled (see FIB-592). The branch `claude/fib-592-lut-validation` is retained, not deleted, so the work and its 14 tests can be picked back up.

PR #464, opened 2026-08-19
[test] Give the overview tab tests a microscope with a fluorescence side (FIB-734)

patrickcleeve2:
Two test files were building a plain Demo session, which is not a compustage — and `DemoMicroscope` only builds a `SimulatedFluorescenceMicroscope` when `sim.is_compustage` is set. So `microscope.fm` was None, and everything followed from that.

| file | before | after |
| -- | -- | -- |
| `test_overview_tab_host.py` | 9 failed, 8 passed, **10 errors** | **27 passed** |
| `test_fm_live_indicator.py` | 5 passed, **5 errors** | **10 passed** |

Across `tests/ui/` and `tests/autolamella/`: **17 failed + 15 errors → 8 failed, 0 errors.**

## One cause, three symptoms

* `build_lamella_poses` returns no fluorescence pose without an FM, and `Lamella.fluorescence_pose`'s setter rejects `None` — which took out **nine tests that have nothing to do with fluorescence**, because the stub window sets a pose on every lamella it adds.
* The fluorescence tab builds no widget, so the `fm_tab` fixture's own assertion failed in setup (**10 errors**, introduced by #452 and extended by #453 — both added tests through that fixture, which was the right thing to write).
* `FMControlWidget` refuses a microscope with no FM outright (**5 errors**).

## The fix

Both fixtures now load `sim-arctis-configuration.yaml` — the same configuration `test_beam_stage_projection.py` already uses for its compustage half, so the three files agree on what an Arctis is rather than each inventing one.

The fixture asserts it got a compustage and an FM rather than trusting the filename, since a stale assumption about the config is precisely what went unnoticed here.

**No test needed changing.** They were written against a microscope the fixture could not supply.

## Verified, not just green

Twenty-four tests went from not running to passing, so "they pass now" is worth very little on its own. I made the fluorescence tab return the *milling* pose instead of its own — the revived tests catch it (3 failures) and previously could not.

## Notes

* Invisible to CI throughout: it installs `.[test]` rather than `.[ui]`, so every test in both files skips there. This is a local-only improvement, which is also why it went unnoticed for a week.
* The eight remaining failures in `test_overview_widget.py` are a **different** problem — overlay shapes and view orientations, no FM involved. Left for their own change rather than folded in here.
* Worth a look separately: `_StubWindow.add_lamella` assigns `build_lamella_poses(...).fluorescence` straight into a setter that rejects `None`, and that function is documented to return `None` on a system without an FM. The production callers may have the same shape.

### DIFF
--- a/tests/ui/test_fm_live_indicator.py
+++ b/tests/ui/test_fm_live_indicator.py
@@ -28,6 +28,7 @@ pytest.importorskip("PyQt5")
 
 from PyQt5.QtWidgets import QApplication
 
+import fibsem.config as _fibsem_config
 from fibsem import utils
 from fibsem.structures import BeamType
 from fibsem.ui.widgets.canvas.quad_view import MicroscopeViewController, QuadViewWidget
@@ -88,6 +89,14 @@ class TestTheFMCanvasCanBeMarkedLive:
         controller.set_fm_live(True)  # must not raise
 
 
+# The simulated compustage, which is the only Demo configuration that has an FM. Shared
+# with tests/ui/test_overview_tab_host.py and test_beam_stage_projection.py so they all
+# agree on what an Arctis is.
+_ARCTIS_CONFIG = os.path.join(
+    os.path.dirname(_fibsem_config.__file__), "config", "sim-arctis-configuration.yaml"
+)
+
+
 class TestTheStreamDrivesIt:
     """The half that was missing. `_update_acquisition_button_states` is the one place
     that already asks `fm.is_streaming` and re-derives the whole panel from it, so the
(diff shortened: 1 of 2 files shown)


patrickcleeve2:
Following up on the last note in this description — I checked, and **the production callers do not have this gap**. Recording it so nobody chases it.

`AutoLamellaFluorescenceOverviewTab` passes `marked_at=FLUORESCENCE_ORIENTATION`, and on that branch `build_lamella_poses` takes the fluorescence position directly:

```python
if marked_at == FLUORESCENCE_ORIENTATION:
    fluorescence_position = deepcopy(state.stage_position)   # never None
else:
    fluorescence_position = _to_fluorescence(...)            # None on an offset system
```

so `fluorescence_position` cannot be None there, and the tab only exists at all when `microscope.fm is not None` (`is_available`). Both conditions on the `return LamellaPoses(..., fluorescence=None)` line are therefore unreachable from that call site. The other production assignment, `structures.py:1586`, is explicitly guarded with `if fluorescence_pose is not None`.

The test stub reached it precisely because it called `build_lamella_poses` *without* `marked_at`, which takes the derived path — so the shape is real, but it was only ever in the stub.

Worth knowing that `_to_fluorescence` returning None on an offset system is deliberate and documented, so a future caller that omits `marked_at` would hit the same thing.

PR #482, opened 2026-08-20
[perf] Blit the tile grid while it is dragged (FIB-752)

patrickcleeve2:
> **Stacked on #481.** Base is `claude/canvas-unreachable-tiles`, so this diff is the blitting alone; GitHub retargets it to `main` when #481 merges.

Dragging the planned grid repainted the **whole canvas** on every motion event — every placed overview image, every overlay — to move some rectangles. `_redraw` ended in `draw_idle`, and a drag emits an event per mouse move.

Per drag frame, 3x3 grid:

| placed overviews | repaint | blitted |
| -- | -- | -- |
| 0 | 6.20 ms | 4.20 ms |
| 6 | 167.35 ms | **4.18 ms** |
| 16 | 427.22 ms | **4.27 ms** |

Flat in the number of images, which is the property worth having — it was the images being re-rendered, not the tiles. Following `PointsOverlay`, which already blits a point drag: mark the tiles `animated`, take one full draw to capture the canvas *without* them, then restore that background and draw only the tiles each frame.

## FIB-627 got here first

[FIB-627](https://linear.app/fibsemos/issue/FIB-627) found this a week earlier and prescribed the same fix; I filed FIB-752 without noticing it. It stays open — it is canvas-level and covers every overlay gesture (the ruler, drag-to-align, the gridbar), while this does the tile grid only, inside the overlay, needing no `CanvasOverlay` protocol change.

It called invalidation "the whole risk", and it was right twice. Both were found by **running the app**, not by the tests.

**A host that repaints mid-drag.** A full draw skips animated artists, so the grid comes off the canvas. On the fluorescence tab that is *every* motion event — it refreshes the stage context and re-declares the world extent from the same handler that moves the grid — which showed as flicker with the grid missing for stretches of the drag. Handled through matplotlib's `draw_event`: any repaint recaptures the background and draws the artists into it.

That handler must **not** call `blit()`. Doing so re-enters the paint, and Qt says so:

```
QWidget::repaint: Recursive repaint detected
QPainter::begin: Paint device returned engine == 0, type: 1
QPainter::end: Painter not active, aborted
```

which is what the first version did. Pinned by a test now.

**A tile landing mid-drag.** Also reachable — the fluorescence tab's `_on_grid_move` has no run guard, unlike the beam tab's `_may_edit_the_plan`. The same handler covers it; verified by swapping the image mid-drag and checking the background recaptures and the grid survives.

## The fluorescence tab was defeating the blit on its own

`set_world_extent` on every motion event asks the canvas to repaint, so two full draws per event survived the blitting and the drag stayed sluggish. Deferred to a new `drag_finished` signal, which arrives once — nobody measures zoom limits mid-gesture. The stage context goes the same way: none of those shapes move when the grid does.

```
FM drag: 33.73 ms/event, 2.0 full canvas draws  ->  4.73 ms/event, 0.1 draws
```

## Failing open

Only while a gesture is in flight; outside one there is nothing to gain and a held background would go stale. Guarded on the canvas offering the calls, and on the blit itself — a host that is not a matplotlib canvas, or a backend that fails part way, falls back to a repaint. Costing a frame's efficiency, not the frame.

Release drops the background and redraws, because artists left `animated` are skipped by every full draw — the grid would otherwise vanish the next time anything repainted, far from the drag that caused it.

## Tests

751 pass. Nine mutations, all killed — one only after a fix: every test called `_on_canvas_drawn` directly, so deleting the `mpl_connect` line entirely passed all 84 overlay tests. There is now a test on the connection itself, which is the thing that would actually regress.

## Not in this PR

A blitted 15x15 is still ~76 ms, and that is now `draw_artist` over 225 individual patches rather than the images — [FIB-753](https://linear.app/fibsemos/issue/FIB-753).

### DIFF
--- a/fibsem/ui/fm/widgets/fm_overview_widget.py
+++ b/fibsem/ui/fm/widgets/fm_overview_widget.py
@@ -274,6 +274,9 @@ class FMOverviewWidget(QWidget):
         # somewhere. None means "wherever the stage is", which is what the runner does
         # by default -- so an untouched grid describes exactly what would be acquired.
         self._target: Optional[FibsemStagePosition] = None
+        # Set only while `_on_drag_finished` is running the refresh a drag deferred.
+        # See there: that refresh must not re-frame the view.
+        self._finishing_drag = False
         self._sparse_selection_enabled = False
         # Where acquired overviews are written. None means nowhere: the widget opens
         # standalone against a simulator as often as it runs inside an experiment, and
@@ -587,6 +590,7 @@ class FMOverviewWidget(QWidget):
         self.tile_grid_overlay.tile_toggled.connect(self._on_tile_toggled)
         self.tile_grid_overlay.grid_resize_requested.connect(self._on_grid_resize)
         self.tile_grid_overlay.grid_move_requested.connect(self._on_grid_move)
+        self.tile_grid_overlay.drag_finished.connect(self._on_drag_finished)
         self.canvas.canvas.canvas_clicked.connect(self._on_canvas_clicked)
         self.canvas.canvas.canvas_double_clicked.connect(self._on_canvas_double_clicked)
         self.canvas.canvas.canvas_right_clicked.connect(self._on_canvas_right_clicked)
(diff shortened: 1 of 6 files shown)


patrickcleeve2:
Closed by GitHub when the base branch `claude/canvas-unreachable-tiles` was deleted on merging #481. Continued as #483 — same branch and commits, rebased onto `main`.

PR #484, opened 2026-08-20
[feature] Run ruff on every pull request

patrickcleeve2:
Nothing has linted this repo before — 689 files, ~233k lines, no linter, no formatter, no style config. This is the first step of a staged ruff rollout tracked in Linear; it deliberately contains only the parts that need no judgement call.

## What's here

Two commits, meant to be reviewed differently.

**`[refactor] Sort imports and drop f-prefixes…`** — machine-generated, 482 fixes across 342 files: 427 unsorted import blocks (`I001`) and 55 f-strings with nothing to interpolate (`F541`). No import is added or removed, no expression changes. Reviewing two files at random is as informative as reviewing all 342.

**`[feature] Run ruff on every pull request`** — the part worth reading: `[tool.ruff]` in `pyproject.toml`, a pinned `dev` extra, and a `lint` job in the workflow. It gets its own job rather than a step inside the test matrix, where it would repeat the same answer eight times. ~10s.

## Enforced now, not after a cleanup

`select` covers the rules that are already clean. The 694 real violations behind everything else are parked in `ignore`, one line per rule with its current count. Burning a rule down means deleting its line — no baseline file, no injected `# noqa`, and the counts make the size of each follow-up visible up front.

Two absences are deliberate, with the reasoning inline in the config:

- **`F401` (unused-import, 395)** cannot be autofixed here, even with a plain `--fix`. Ruff classes most of those removals as *safe*, but a dozen modules exist only to re-export under a shorter name, and ruff honours `X as X` as an explicit re-export while a *renaming* alias reads as unused. Running the fix empties `fibsem/applications/autolamella/config.py` to zero bytes and `import fibsem.applications.autolamella.structures` then dies with `AttributeError: ... has no attribute 'EXPERIMENT_NAME'`. Giving those modules an explicit `__all__` is the prerequisite, and it's its own piece of work.
- **`RUF100` (unused-noqa)** is evaluated relative to `select`, so with a narrow rule set it deletes the suppressions for rules that aren't enabled *yet*. Enabling it stripped 133 `# noqa` comments — 95 of them `E402`, a rule ignored in this same config, so every one would need retyping later — along with the comments explaining why they were there. It belongs at the end of the ratchet.

Notebooks are excluded: ruff rewrites `.ipynb` in place, re-sorting imports inside the cell JSON. Whether they're in scope is a separate question.

`target-version` is left unset on purpose — ruff reads `requires-python` and infers py38, so the two can't drift.

## Verification

- `ruff check .` — clean.
- Full test suite: **5,680 passed, 24 skipped, 0 failed** (105s). That's everything, including the `ui`-marked tests CI skips.
- `# noqa` count unchanged at 133 — the 23 appearing in the diff are moved by import sorting, not deleted.

## Next

`F401` burn-down (blocked on the `__all__` decision), then the small real-bug rules — `F821` (24, includes at least one genuine `NameError`), `F811`, `F841`, `F402`, `F601` — then `ruff format`, which is a separate conversation because it rewrites 629 files.


patrickcleeve2:
### Machine-checking the "it only reorders imports" claim

A 342-file tool-generated diff is not reviewable by reading it, so the claim it rests on was checked mechanically instead. For every touched file, the version on `main` and the version here were parsed and compared on three axes:

1. **Logic** — strip all imports, fold placeholder-free f-strings to plain strings, compare the dumped ASTs.
2. **Bindings** — compare the set of `(module, level, name, alias)` tuples each file imports.
3. **Crossing** — compare which gap between non-import statements each import falls into, so an import moving past a real statement (e.g. over a `matplotlib.use()` call) would show up.

```
files checked ............................. 344
files where logic changed ................. 0
files where an import binding changed ..... 0
files where an import crossed a statement . 0
files where a duplicate import collapsed .. 6
```

The last row is the one caveat: `I001` collapsed 9 exact duplicate imports across 6 files — `import glob` twice in `fibsem/segmentation/_nnunet.py`, `Qt` and `pyqtSignal` twice in `fibsem/ui/FibsemSpotBurnWidget.py`, `FunctionWorker` twice in `fibsem/applications/autolamella/ui/AutoLamellaUI.py`, plus three more. Same module, same name, same alias, so the second import was rebinding the same object.

**What this does not cover:** order *within* a single contiguous import block. If one module must be imported before another for its import-time side effects and both sit in the same block, ruff reorders them and no AST comparison would catch it. That class of failure surfaces at import time rather than at runtime, and importing is the first thing the suite does — 5,715 tests across 8 jobs, both Windows runners included.

<details><summary>the checker</summary>

```python
import ast, subprocess
BASE = "2208a261"
files = subprocess.run(["git","diff","--name-only",BASE,"HEAD","--","*.py"],
                       capture_output=True,text=True).stdout.split()
def blob(r,p): return subprocess.run(["git","show",f"{r}:{p}"],capture_output=True,text=True).stdout
def parts(src):
    tree, imps, gaps, gap = ast.parse(src), [], [], 0
    for n in tree.body:
        if isinstance(n,(ast.Import,ast.ImportFrom)):
            for a in n.names:
                b=(getattr(n,"module",None),getattr(n,"level",0),a.name,a.asname)
                imps.append(b); gaps.append((gap,)+b)
        else: gap += 1
    class N(ast.NodeTransformer):
        def visit_Import(self,n): return None
        def visit_ImportFrom(self,n): return None
        def visit_JoinedStr(self,n):
            if not any(isinstance(v,ast.FormattedValue) for v in n.values):
                return ast.Constant(value="".join(v.value for v in n.values if isinstance(v,ast.Constant)))
            return n
    return set(imps), imps, set(gaps), ast.dump(ast.fix_missing_locations(N().visit(ast.parse(src))))

for f in files:
    (sb,lb,gb,kb),(sa,la,ga,ka) = parts(blob(BASE,f)), parts(blob("HEAD",f))
    if kb != ka: print("LOGIC CHANGED:", f)
    if sb != sa: print("BINDING CHANGED:", f)
    if gb != ga: print("IMPORT CROSSED A STATEMENT:", f)
    if len(lb) != len(la): print("duplicate collapsed:", f, len(lb)-len(la))
```
</details>

PR #488, opened 2026-08-21
[refactor] Format the tree with ruff format

patrickcleeve2:
The formatter flag day. 599 files reformatted, whitespace only, plus a CI step so it stays that way.

**Do not review this file by file.** It is machine-generated and no human can usefully read 25k lines of whitespace. Review the two small commits and the proof below instead.

## Commits

| | |
|---|---|
| `bc08bbc6` | `ruff format .` — 599 files, nothing else in the commit |
| `e7131258` | `ruff format --check .` in the lint job, and `**/*.md` excluded |
| (third) | `.git-blame-ignore-revs`, with the header but no SHA yet — see below |

## Proof

Formatting changes no tokens, so the invariant is stronger than a code review: **every file must parse to an identical AST**. Checked across all 599:

```
files reformatted ................ 599
files whose AST changed .......... 0
files that failed to parse ....... 0
```

Docstrings are compared with per-line whitespace collapsed, because re-indenting a docstring genuinely does change the string's value. Nothing else needed normalising — quote-style changes (`'x'` → `"x"`) alter the token but not the AST value.

Full suite: **5,743 passed, 24 skipped**. That is the same count as `main`.

## Markdown is excluded, deliberately

Ruff formats Python inside fenced code blocks, and it wanted to change six documentation files. Documentation snippets are written to be read: it wrapped a three-argument `setup_session(...)` call in `README.md` across three lines, and rewrote 84 lines of `docs/design/autofocus.md`. Bundling that into a whole-tree code reformat guarantees nobody reviews it. `[tool.ruff.format] exclude = ["**/*.md"]` keeps the docs question open rather than answering it by accident.

## What this costs you, and the recipe

There are 11 open pull requests. Every one will conflict, because this touches nearly every line of nearly every file. **Do not resolve those conflicts by hand, and do not use `merge -X ours`** — most hunks conflict, so that silently drops real changes from `main`.

For each branch:

```bash
git rebase <the commit before the format commit>   # your branch, unformatted, on an unformatted base
ruff format .                                       # now format your branch the same way
git commit -am "format"
git rebase main                                     # both sides formatted; conflicts evaporate
```

The conflicts disappear because both sides are then the output of the same deterministic formatter.

## `.git-blame-ignore-revs`

The file lands here with its header and no SHA, because this repository squash-merges: the revision to record is the one that ends up on `main`, which does not exist until this is merged. A one-line follow-up adds it immediately after.

Once merged, each clone wants this once:

```bash
git config blame.ignoreRevsFile .git-blame-ignore-revs
```

GitHub reads the file on its own.

## Recommended before merging

Launch the app. The suite covers a great deal, but a formatter interacting badly with something Qt-related is exactly the class of thing that passes tests and fails on screen, and the app is the check that caught nothing on the import-sort PR precisely because it is the one that could have.


patrickcleeve2:
Superseded by the incremental approach: formatting the files each pull request touches, rather than the whole tree at once. A 599-file commit is a lot to absorb in one go, and it would have forced a rebase on all eleven open pull requests.

Worth recording what this branch established, since it stays true whenever the flag day is reconsidered: 599 files reformat, **zero ASTs change**, and the full suite is unaffected at 5,743 passed. The AST checker is in a comment on #484. Also found here: ruff formats Python inside fenced code blocks in Markdown, and wanted to rewrite six documentation files — the replacement excludes `**/*.md` for that reason.

The branch is left in place rather than deleted, in case the flag day is wanted later.

patrickcleeve2:
For the record before this goes cold: CI was **fully green** on this branch — all 9 checks, including both Windows runners and Python 3.8 through 3.13. So the whole-tree reformat is verified end to end, not just locally, and the branch is one merge away if the flag day is ever preferred to the incremental route in #489.

PR #496, opened 2026-08-21
[refactor] Remove unused imports from imaging, milling, fm, segmentation, alignment

patrickcleeve2:
First slice of the `F401` burn-down. 50 unused imports removed across 20 files. **F401 goes 389 → 339** repo-wide; in these five areas, 82 → 32, and all 32 that remain are in `__init__.py`, where the imports *are* the re-export idiom and ruff correctly refuses to touch them.

## `ruff --fix` alone is not safe here, and here is the proof

Ruff answers "is this name used inside this file". That is a different question from "is it safe to delete". Every flagged name was checked against the whole repo, three ways, **before** touching anything:

| reachable how | findings at risk |
|---|---|
| `from M import X` elsewhere | **15** |
| `import M` then `M.X` | 0 |
| `from M import *` | 2 (nobody actually imports them) |

Fourteen of the fifteen are in `__init__.py`, which ruff leaves alone. **One is not:** `fibsem/microscopes/tescan.py` re-exports `tescanautomation`, [scripts/test_installation.py:60](https://github.com/fibsem-os/fibsem-os/blob/main/scripts/test_installation.py#L60) imports it, and **ruff rates that removal `safe`**. So `ruff check --fix` on `main` today silently breaks that script. None of the fifteen is in this slice; the tescan one is flagged for whichever slice reaches `fibsem/microscopes`.

Then the same check ran **after** the edit, against what was actually removed: zero of the 50 are imported anywhere else.

## The check still had a hole, and the suite found it

`tests/fm/test_acquisition.py` asserted against `acquisition.PREVIEW_MAX_DIMENSION` — a constant that lives in `fibsem.imaging.reduce` and was reachable there only because `acquisition.py` imported it and never used it. Removing the import broke the test.

That is a **fourth** path, and the one my pre-checks missed: `from fibsem.fm import acquisition` binds a *module* via `ImportFrom`, so it never appears in an `import M` alias map, and `acquisition.PREVIEW_MAX_DIMENSION` reaches straight through it.

Detector fixed and re-run over the remaining 339: one other instance, `fibsem.milling.setup_milling` used by `example/example_milling.py`, in an `__init__.py` ruff will not auto-fix anyway.

Fixed by pointing the test at the constant's real home, rather than restoring an accidental re-export.

## Why these five areas

Reformatting a file is mandatory once it is touched, and the cost per finding spans two orders of magnitude:

| area | findings | format lines | lines per finding |
|---|---|---|---|
| `fibsem/imaging` | 18 | 165 | **9** |
| `fibsem/milling` | 18 | 345 | 19 |
| `fibsem/fm` | 27 | 405 | 15 |
| `fibsem/segmentation` | 15 | 474 | 32 |
| `fibsem/alignment` | 4 | 48 | 12 |
| — for contrast — | | | |
| `fibsem/structures.py` | 3 | 1,045 | **348** |
| `fibsem/microscope.py` | 11 | 1,364 | 124 |

Doing all 146 files at once would be an 18,406-line diff. This is the cheap end.

## Commits

- `a638b2c2` — the 50 removals, nothing else
- `8dd1af5f` — the mandatory reformat, plus the one-line test repoint, called out in the message rather than buried

## Verification

`ruff check .` clean, all touched files formatted, full suite **5,752 passed, 24 skipped**.

### DIFF
--- a/fibsem/alignment/plotting.py
+++ b/fibsem/alignment/plotting.py
@@ -11,7 +11,6 @@ from fibsem.structures import ImageSettings
 
 if TYPE_CHECKING:
     from matplotlib.axes import Axes
-    from matplotlib.figure import Figure
 
     from fibsem.alignment import AlignmentDifferential, AlignmentIteration
     from fibsem.structures import FibsemImage
@@ -29,7 +28,6 @@ def _plot_image_with_crosshair(ax: Axes, data: np.ndarray, title: str) -> None:
 
 def _alignment_save_path(ref_image: FibsemImage) -> tuple:
     """Return (ref_path, prefix, ts) for saving alignment plots."""
-    from datetime import datetime
 
     from fibsem.alignment import ALIGNMENT_SUBDIR
 
--- a/fibsem/alignment/plotting.py
+++ b/fibsem/alignment/plotting.py
@@ -98,15 +98,23 @@ def plot_multi_step_alignment(
         dy_px = r.shift.y / pixel_size
         colour = "lime" if r.score >= 0.7 else ("orange" if r.score >= 0.5 else "red")
         axes[0, 1 + i].text(
-            0.04, 0.04,
+            0.04,
+            0.04,
             f"dx={dx_px:.1f}px  dy={dy_px:.1f}px",
             transform=axes[0, 1 + i].transAxes,
-            color=colour, fontsize=7, va="bottom", fontfamily="monospace",
-            bbox=dict(boxstyle="round,pad=0.2", facecolor="black", alpha=0.6, edgecolor="none"),
+            color=colour,
+            fontsize=7,
+            va="bottom",
+            fontfamily="monospace",
+            bbox=dict(
+                boxstyle="round,pad=0.2", facecolor="black", alpha=0.6, edgecolor="none"
+            ),
         )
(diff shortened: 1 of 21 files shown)


patrickcleeve2:
Closing on review feedback: 21 files is unreadable, regardless of how it was verified. Re-slicing into five PRs of at most 5 files each, one coherent module per PR, opened independently off `main` so they can merge in any order.

The findings from this branch carry over to those PRs and are worth not losing:

- **`ruff --fix` alone is unsafe here.** 15 flagged names are imported elsewhere. Fourteen are in `__init__.py`, which ruff leaves alone; **one is not** — `fibsem/microscopes/tescan.py` re-exports `tescanautomation`, `scripts/test_installation.py:60` imports it, and ruff rates that removal **safe**. Flagged for whichever slice reaches `fibsem/microscopes`.
- **A fourth transitive path exists that static import analysis misses:** `from pkg import mod` binds a *module* via `ImportFrom`, so `mod.X` reaches through it without ever appearing in an `import M` alias map. That is how removing an unused import in `fibsem/fm/acquisition.py` broke `tests/fm/test_acquisition.py`. Only the test suite caught it.
- **Cost per finding spans two orders of magnitude** under format-on-touch: `fibsem/imaging` is 9 format-lines per unused import, `fibsem/structures.py` is 348.

PR #531, opened 2026-08-24
[refactor] Type tiled acquisition progress events (FIB-402)

patrickcleeve2:
## Summary

- replace tiled acquisition dictionaries with a frozen, keyword-only tagged event model
- type the shared psygnal contract and all beam/FM producers
- migrate the two beam overviews, FM overview, main status bar, and queued Qt bridges to class/attribute dispatch
- remove dictionary compatibility and the unused tileset task tag
- preserve existing wording, counts, ETA, event ordering, previews, cancellation, stitching, saving, and terminal behavior; FIB-742 presentation remains deferred

## Verification

- 166 passed, 1 deselected: beam runner, FM acquisition, and typed payload suites (the deselected OME schema test requires network access)
- 479 passed: four tiled UI consumer suites
- 381 passed, 1 skipped: broader tiling/import/isolation suites
- Ruff and git diff checks pass

### DIFF
--- a/fibsem/applications/autolamella/ui/AutoLamellaMainUI.py
+++ b/fibsem/applications/autolamella/ui/AutoLamellaMainUI.py
@@ -80,6 +80,16 @@ from fibsem.applications.autolamella.workflows.workflow_estimate import (
     estimate_addition,
     estimate_workflow,
 )
+from fibsem.imaging.tiling.progress import (
+    BeamTileCompletedEvent,
+    CountedTiledPhaseEvent,
+    CountedTiledTerminalEvent,
+    FluorescenceTileCompletedEvent,
+    FluorescenceTileCountEvent,
+    TiledAcquisitionEvent,
+    TiledOutcome,
+    TiledTerminalEvent,
+)
 from fibsem.structures import BeamType
 from fibsem.ui import notification_service
 from fibsem.ui.FibsemMinimapWidget import FibsemMinimapWidget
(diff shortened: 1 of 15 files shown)


patrickcleeve2:
Superseded — closing.

The work this proposed is done, landed as a four-PR stack under FIB-402 plus three more under FIB-401:

| PR | |
| -- | -- |
| [#547](https://github.com/fibsem-os/fibsem-os/pull/547) | format prep |
| [#548](https://github.com/fibsem-os/fibsem-os/pull/548) | the `TiledProgress` contract |
| [#549](https://github.com/fibsem-os/fibsem-os/pull/549) | consumers |
| [#550](https://github.com/fibsem-os/fibsem-os/pull/550) | producers |
| [#554](https://github.com/fibsem-os/fibsem-os/pull/554)–[#556](https://github.com/fibsem-os/fibsem-os/pull/556) | the same treatment for `acquisition_progress_signal` (FIB-401) |

**What changed from this design.** Review here settled on a single flat frozen record rather than a tagged union, replacing the 8 classes, 3 enums, `Union` alias and 2 helpers with one `TiledProgress` and one `TiledStatus`. The argument was that every consumer dispatch reduced to `isinstance(event, counted_types) and event.total` — the `isinstance` was ceremony, and the presence check was doing the work. With one variant per producer and every consumer internal, the hierarchy bought type narrowing nobody used.

Three things this PR got right that carried over unchanged: the frozen dataclass, the terminal-status question answered once on the type rather than restated at each consumer, and 0-based indices on the wire with the 1-based form derived for display.

Also carried over, and worth recording: `eq=False` is load-bearing. A generated `__eq__` compares `preview`, which holds an image whose numpy data compares elementwise, raising `ValueError` from inside a Qt slot — but only for two *distinct* events, since comparing one to itself short-circuits on identity. The obvious test passes while the real case raises.

`message` was dropped from the wire entirely: each consumer now owns its own wording via a `_STATUS_LABELS` table, so the same report reads differently in the minimap and the status bar without the producer choosing for them.

PR #543, opened 2026-08-25
[fix] Skip cleanly when the autofocus sweep is declined, instead of crashing on None

patrickcleeve2:
## What

`run_auto_focus` returns `None` on backends where the working distance is not settable for the requested beam (TESCAN ION — the #541 gate), but `_run_autofocus` in tasks/base.py passed the result straight to `result.save(...)` → AttributeError that kills the task the first time a Tescan protocol enables autofocus (`select_position` and `spot_burn` both run the ION leg). Config-gated, off by default — but tonight's hardware session shouldn't have a loaded footgun.

Guards at the caller: logs an explicit **skipped** status (visible in the workflow status UI) and saves nothing.

## Why not an empty result instead

Considered and rejected: `AutoFocusResult` requires a working distance, a focus score, and an image; `save()` writes a completed-looking artifact directory (`data.json`/`best.tif`/`plot.png`); and the next status line logs `Autofocus complete: WD=…` — a placeholder result fabricates all three for a sweep that never ran, which is exactly the fake success the #541 gate removed. `None` means nothing happened; the caller now says so.

## Verification

- 2 new tests in `tests/autolamella/test_spot_burn_autofocus.py` (the file that owns `_run_autofocus` coverage): the skip path (no raise, 'skipped' status, nothing written) and the guard-doesn't-swallow-results path. 16/16 pass.
- Mutation: disabling the guard fails exactly the skip test (AttributeError), everything else stays green.
- `ruff check` clean; both files carry **pre-existing** format dirt (dirty at HEAD before this change) — deliberately left for the tescan→main sync formatting pass rather than churning a hot core file here.

### DIFF
--- a/fibsem/applications/autolamella/workflows/tasks/base.py
+++ b/fibsem/applications/autolamella/workflows/tasks/base.py
@@ -504,6 +504,17 @@ class AutoLamellaTask(ABC):
             settings=settings,
             stop_event=self._stop_event,
         )
+        if result is None:
+            # run_auto_focus declined the sweep (the working distance is not settable
+            # for this beam on this backend -- TESCAN ION, see FIB-508). Say "skipped"
+            # and save nothing: a placeholder result would write a completed-looking
+            # artifact directory and log a working distance that was never applied.
+            self.log_status_message(
+                "AUTOFOCUS",
+                f"Autofocus skipped: the {beam_type.name} working distance is not settable on this system.",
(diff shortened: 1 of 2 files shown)


patrickcleeve2:
Superseded by #544 targeting main — the #541 gate this fixes lives on main, so the guard belongs there too (per review). The tescan branch picks it up on the next main sync.

PR #560, opened 2026-08-26
[feat] Spot burn preset selection for preset-driven backends (TESCAN)

patrickcleeve2:
## What

On TESCAN the spot burn conditions were a lie: the widget offered a Beam Current combo (fed from a hardcoded fake current list) that `run_spot_burn` then ignored in favour of a hardcoded module constant (`SPOT_BURN_PRESET = "30 keV; 100 pA"`). The preset is now a real, user-visible setting.

- `SpotBurnSettings` gains an optional `preset` (serialized; `None` = backend default; `milling_current` stays ignored wherever a preset drives the beam).
- `TescanMicroscope.run_spot_burn` burns at `settings.preset`, falling back to `SPOT_BURN_PRESET` when unset. `_create_spot_burn_layer` takes the preset as a parameter.
- The spot burn tab shows a **Preset** combo on TESCAN — fed from `get_available_values_cached` (building a tab must not query the microscope; the Beam Current path now uses the cached getter too) — and keeps the Beam Current combo on every other backend. Off-list protocol presets stay exactly selectable so an untouched value round-trips losslessly (mirrors the off-grid current rule).
- The workflow task config carries the preset. It is deliberately excluded from the generic parameter form (like `coordinates`): it's chosen in the spot burn tab's combo, not as free text. Pre-preset protocols load as `None`.

Preset names are free-form on the instrument (e.g. "30 keV; 100 pA; my cool preset"), so nothing parses them here.

## Verification

- 61 tests green across `test_tescan_spot_burn.py` (new: settings preset reaches the DrawBeam layer), `tests/autolamella/test_spot_burn.py` (new: preset round-trips, pre-preset protocols default to None, preset stays out of the parameter form), the new `tests/test_spot_burn_widget_preset.py` (Tescan gets the preset combo from the seeded cache, non-Tescan keeps the current combo, off-list presets round-trip), and the coordinates-widget suite.
- ruff check clean; the four pre-format-dirty files got a separate whole-file format commit, proved AST-identical (docstring trailing whitespace normalised).

## Notes

- Second commit is format-only.
- Not yet exercised on hardware; the preset reaches the same IEtching field the hardware-verified default did, so risk is confined to the plumbing covered by tests.

### DIFF
--- a/fibsem/applications/autolamella/workflows/tasks/spot_burn.py
+++ b/fibsem/applications/autolamella/workflows/tasks/spot_burn.py
@@ -56,10 +56,15 @@ class SpotBurnFiducialTaskConfig(AutoLamellaTaskConfig):
         default_factory=list,
         metadata=field_meta(tooltip="Spot burn positions in normalised image coordinates (0-1)"),
     )
+    # beam preset for preset-driven backends (TESCAN); None means the backend default.
+    # Chosen in the spot burn tab's preset combo, not the generic parameter form.
+    preset: Optional[str] = None
 
     @property
     def parameters(self) -> tuple[str, ...]:
-        return tuple(p for p in super().parameters if p != "coordinates")
+        # coordinates are edited on the canvas, the preset in the spot burn tab's
+        # combo -- neither belongs in the generic parameter form
+        return tuple(p for p in super().parameters if p not in ("coordinates", "preset"))
 
     @property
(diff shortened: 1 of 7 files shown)


patrickcleeve2:
Shelved for now — not merging changes to the spot burn path at this time. The branch (claude/tescan-spot-burn-preset) is kept; reopen or rebase when the feature is wanted.

PR #562, opened 2026-08-26
[feat] FibsemMillingSettings.preset defaults to None; TESCAN hard-errors on a missing preset

patrickcleeve2:
**Stacked on #561** (it updates the estimation-model comment that PR adds) — GitHub will retarget to main when #561 merges. Stacked PRs get no CI here; local gate: 322 tests green across `tests/milling`, the tescan lifecycle/spot-burn/estimate suites, the form-builder and stage-list widget suites; ruff check + format clean.

## What

`preset` now means what it says: **None = not set** (ThermoFisher, simulator — no more fake-looking `"30 keV; 2nA"` on every stage), **string = chosen**. On TESCAN, `setup_milling` raises a clear `ValueError` **before touching the hardware** when a stage has no preset — inheriting whatever preset happens to be active would be silent condition-dependence.

## What it does NOT do

It does not make preset-presence imply TESCAN retroactively: `to_dict` always wrote the preset, so protocols saved before this change carry the old default string explicitly, on every backend. That is why the estimation-model switch in #561 stays a driver registration rather than preset inference — the comment there is updated to say so. Old protocols on either backend load unchanged (covered by tests).

## None-safety in the readers

- stage-list preset combo and the form builder's `items: "dynamic"` combo render an **empty** combo for an unset preset, not a literal "None" entry
- the pattern-plot label skips an unset preset
- everything else already tolerated None (`ValueComboBox.set_value(None)` is a no-op on string combos; the estimate parser returns None)

## Tests

- default is None; None and string presets round-trip; an old protocol dict with the explicit string loads unchanged
- `setup_milling` with `preset=None` raises and nothing reaches the fake connection (no layer, no set calls)
- second commit is a whole-file format pass on the four pre-format-dirty files, proved AST-identical

## Update: system-config fallback (out-of-the-box behaviour)

Review question: do stock protocols still work on TESCAN? They carry **no** preset keys, so under the original hard error they would have refused to mill. `setup_milling` now resolves **stage preset → machine config's default milling preset → hard error**, with an info log when the fallback fires. The config's `milling: preset` field ([USER]-marked) existed but was consumed nowhere — the drivers only receive `SystemSettings`, so that now carries a read-only copy of the config milling block (deliberately not emitted by `to_dict`; the canonical home stays `MicroscopeSettings.milling`). The shipped tescan config default is set to `"30 keV; 1nA"`.

Net behaviour:
- existing app-saved protocols: unchanged (they carry presets explicitly)
- stock/hand-written protocols without presets: mill at the site's configured default, logged
- no stage preset **and** no config default: hard error before touching hardware

New tests: fallback fires, stage preset wins over the default, refusal only when neither is set.

### DIFF
--- a/fibsem/config/tescan-configuration.yaml
+++ b/fibsem/config/tescan-configuration.yaml
@@ -59,4 +59,4 @@ milling:
     rate:                       1.3e-08                     # the cryo ion etching rate         (TESCAN)          [m3/A/s]  [USER]
     spacing:                    0.005                       # the cryo exposition mesh spacing  (TESCAN)      [dimensionless][USER]
     spot_size:                  1.0e-06                     # the cryo milling spot size        (TESCAN)            [metres][USER]
-    preset:                     "30 keV; 20 nA"             # the default milling preset        (TESCAN)                    [USER]
+    preset:                     "30 keV; 1nA"               # the default milling preset        (TESCAN)                    [USER]
--- a/fibsem/microscopes/tescan.py
+++ b/fibsem/microscopes/tescan.py
@@ -1273,6 +1273,15 @@ class TescanMicroscope(FibsemMicroscope):
         if mill_settings.milling_channel is not BeamType.ION:
             raise ValueError("Only FIB milling is currently supported.")
 
+        # fail before touching the hardware: TESCAN milling is preset-driven, so a
+        # stage without a preset has no beam conditions -- inheriting whatever
+        # preset happens to be active would be silent condition-dependence
+        if mill_settings.preset is None:
+            raise ValueError(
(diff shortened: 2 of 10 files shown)


patrickcleeve2:
Closing: this can't serve the estimates (old protocols carry the preset string explicitly on every backend, so preset-presence can never be inferred retroactively — the driver registration in #561 stands alone), which leaves only a hygiene benefit that doesn't justify the blast radius (FibsemMillingSettings default change + a SystemSettings schema field with an asymmetric serialization invariant). If fail-loud-on-missing-preset is wanted later, it can be a tiny standalone change in TescanMicroscope.setup_milling.

PR #591, opened 2026-08-27
[refactor] Delete movement_progress_signal, which duplicated ensure_main_thread (FIB-825)

patrickcleeve2:
FIB-402 has this signal on its list to be *typed*. Looking at what it is **for** rather than what it carries turned that into a deletion.

## It was a thread hop, and the widget already had one

Seven emits in `@thread_worker` bodies, one consumer on the same class, writing a `QLabel`. Something must marshal that — but this widget already marshals a different way, in the same file, **on the line adjacent to an emit**:

```python
self.movement_progress_signal.emit({"msg": ACQUIRING_IMAGES})   # via the signal
self.update_ui_after_movement()                                  # @ensure_main_thread, direct
```

`update_ui` (`:445`) and `update_ui_after_movement` (`:468`) were already decorated. That decorator is the house contract — 23 sites across `fibsem/ui/` — and `FMOverviewWidget` says so in a comment: relaying through a signal is *"this widget's own dialect"*.

**And one emit marshalled nothing at all.** `move_stage_finished` is connected to `worker.finished`, which Qt delivers on the GUI thread, so its `{"finished": True}` was a GUI-thread → signal → GUI-thread round trip.

## The change

`_set_move_status` gains the decorator; the `finished` branch becomes `_on_move_finished`; the seven emits become seven direct calls that say what they mean. The declaration, the `connect`, the dict-parsing handler and the dict all go.

**No new types.** Typing this payload would have been ceremony — `msg` is producer-authored (`f"Moving to {stage_position.pretty}…"`), so the "consumer owns the wording" argument that justified dropping `message` from the tiled and fluorescence signals never applied here.

## Semantics — checked, not assumed

| | |
| -- | -- |
| cross-thread | `superqt.ensure_main_thread` is non-blocking by default; a cross-thread Qt emit is queued too. Both deliver through the main-thread event loop **in call order**, so the message sequence a move produces is unchanged. |
| same-thread | becomes a direct call — affects only the round trip that was already pointless. |
| `await_return=True` | **not used.** Blocking a worker on the GUI thread is how a stage move becomes a freeze. |

## Tests

They used the signal as an *observation hook*, so they now hook the controller's `set_info` instead — a truer observation, since it samples the value that actually reached the bar at the moment it reached it, on the thread that wrote it.

`test_the_status_is_written_on_the_gui_thread` keeps its assertion and gets a new reason: the invariant used to hold because a queued signal delivered onto the GUI thread, and now holds because of the decorator. That is the one thing this change could plausibly have broken, so it stays asserted rather than dropped.

**17 passed** (`test_movement_progress.py`, `test_movement_widget_vertical_gate.py`), plus **40** in `test_viewer_less_widgets.py` / `test_stage_frame.py`. Decorators AST-checked onto the right methods. `ruff check` and `ruff format --check` clean.

Note `tests/ui/` does not run in CI (the `[ui]` extra is deliberately not installed), so those 17 were run locally with `QT_QPA_PLATFORM=offscreen`.

### DIFF
--- a/fibsem/ui/FibsemMovementWidget.py
+++ b/fibsem/ui/FibsemMovementWidget.py
@@ -37,8 +37,6 @@ ACQUIRING_IMAGES = "Acquiring images…"
 
 
 class FibsemMovementWidget(QtWidgets.QWidget):
-    movement_progress_signal = QtCore.pyqtSignal(dict)
-
     def __init__(
         self,
         microscope: FibsemMicroscope,
@@ -233,7 +231,6 @@ class FibsemMovementWidget(QtWidgets.QWidget):
         self.saved_positions_widget.move_to_requested.connect(self.move_to_position)
 
         # signals
-        self.movement_progress_signal.connect(self.handle_movement_progress_update)
         self.image_widget.acquisition_progress_signal.connect(
             self.handle_acquisition_update
         )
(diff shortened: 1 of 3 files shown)


patrickcleeve2:
Shelved as a draft, not abandoned — the work is complete and green.

Held pending [FIB-828](https://linear.app/fibsemos/issue/FIB-828), which asks the better question underneath this one: `absolute_movement_worker` calls the widget's own GUI methods, and the real fix may be to lift the movement operation out of the widget entirely.

That matters here because it could invert the conclusion. This PR argues the signal is ceremony **because the producer is the widget itself** — a widget signalling itself. If the operation moves out, the producer becomes a non-UI object, and a non-UI producer reporting progress is exactly the case where a signal is the right mechanism rather than ceremony.

The signal would not return unchanged either way: `{msg}` / `{finished}` is shaped around this widget's needs, and an extracted service would want its own shape. But that decision belongs in FIB-828.

To be clear about what is *not* the reason: the current calls are **safe**. `_set_move_status` and `update_ui_after_movement` both carry `@ensure_main_thread`, so they are marshalled onto the GUI thread. The concern is coupling, not threading.

patrickcleeve2:
Superseded by #617.

This PR's premise no longer describes the code. It argued `movement_progress_signal` was a thread hop duplicating `@ensure_main_thread` — true when written, but #612 moved the movement reporting onto the GUI thread, so nothing emits from a worker any more and there is no hop to remove.

The deletion is still right, for a simpler reason: the signal is a widget calling itself, synchronously, with one subscriber and no external users. Measured rather than assumed — every delivery is a direct connection, with the handler running inside `emit()`.

Not rebased because the diff (+79 −74) is against `absolute_movement_worker` as it stood two rewrites ago — #612 and #616 have both been through this file since. #617 redoes it against current main and carries the test seam across.

Closing in favour of that.

PR #594, opened 2026-08-27
[refactor] Declare a typed contract for workflow status updates (FIB-827)

patrickcleeve2:
The `status` value nested inside `workflow_update_signal`’s payload is a bare 13-key dict, read through eleven separate `.get()` calls across two consumers.

## Scope: the nested record only, deliberately

The envelope stays exactly as it is — `{"msg": …, "status": …}`. Most of that signal is **not a signal at all**: ten of its thirteen emits are blocking calls built from a shared mutable bool and a polling loop, and untangling that is a threading redesign ([FIB-826](https://linear.app/fibsemos/issue/FIB-826)). This record is independent of that work **in both directions** and can land before or after it.

## Why it is the easy half

There was no vocabulary to design. `status` has always been a real `AutoLamellaTaskStatus` on the wire — unlike the other four signals in FIB-402’s scope, the discriminator was already correct. What was missing was a shape around it.

Nothing emits one yet: both consumers decode through `WorkflowStatusUpdate.from_payload`, which accepts the dict the producer still sends and the record it will send next. **One decode at the boundary** rather than a `.get()` per field.

## `from_payload` is total by construction, and that is the point

It runs inside a queued Qt slot, where PyQt5 turns any escaping exception into `qFatal` — the process aborts mid-run and the abort reaches **no logfile** ([FIB-329](https://linear.app/fibsemos/issue/FIB-329)). This signal has already killed the app that way once, over a missing `msg` key; `tests/ui/test_workflow_update_payload.py` exists because of it.

So every field has a default, nothing indexes, and an unrecognised `status` becomes `NotStarted` — a value chosen because it fires no consumer branch, so it renders "nothing in particular happened" rather than crashing or picking a wrong outcome.

## Three decisions worth reviewing

**`eq=False`, for a different reason than `TiledProgress`’s.** Three fields hold lists and `WorkItem` is a *mutable* dataclass, so it is unhashable; `frozen=True` with the default `eq=True` generates a `__hash__` that raises `TypeError`. No numpy involved. Pinned by a test that *also* asserts `WorkItem` really is unhashable, so the reason cannot rot into a stale comment.

**`lamella_name` becomes a property, not a field** — one source of truth, so the deprecated alias cannot drift from `item_name`, and it deletes cleanly with the HookContext shims after v0.6.

**`queue_position` stays 1-based**, which is what the producer already writes. Deliberately unlike milling’s 0-based `current_stage` that seven consumers each increment — porting a display helper here would double-count.

## One behaviour difference, closed deliberately

`queue_total` now defaults to `0` rather than `None`, so the status-bar guard reads truthiness instead of `is not None`. A count of 0 means "nothing to be in the middle of", and the old guard would have rendered `"3/0"` for a malformed payload.

**607 passed** across `tests/autolamella/` plus the timeline and payload-safety suites. `tests/autolamella/` runs in CI. Static 3.8 scan clean.

### DIFF
--- a/fibsem/applications/autolamella/ui/AutoLamellaMainUI.py
+++ b/fibsem/applications/autolamella/ui/AutoLamellaMainUI.py
@@ -71,6 +71,9 @@ from fibsem.applications.autolamella.ui.workflow_timeline_widget import (
     WorkflowProgressWidget,
 )
 from fibsem.applications.autolamella.workflows.tasks.queue import QueueOp, QueueResult
+from fibsem.applications.autolamella.workflows.tasks.status import (
+    WorkflowStatusUpdate,
+)
 from fibsem.applications.autolamella.workflows.tasks.tasks import get_task_supervision
 from fibsem.applications.autolamella.workflows.workflow_estimate import (
     AdditionEstimate,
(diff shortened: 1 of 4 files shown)


patrickcleeve2:
Pushed three fixes after a second look — one of them was a real behaviour change I had not flagged.

## The real one: absent vs empty queue snapshot

`update_from_status` guarded on `queue_items is None`. My record defaulted the field to `[]`, so that guard could never fire, and I had quietly weakened it to `if not queue_items`. That collapses two states the dict kept apart:

| payload | means | old behaviour |
| -- | -- | -- |
| no `queue_items` key | this payload says nothing about the queue | leave the rows alone |
| `queue_items: []` | the queue is empty | reconcile the timeline to **zero rows** |

With the falsy check, **an emptied queue would leave stale rows on screen**. The field is now `Optional[List[WorkItem]] = None` and `from_payload` preserves the difference. Two tests pin it, plus a guard asserting the consumer keeps checking `is None`.

## Exactness

`_active_start_time` now uses `timestamp if timestamp is not None else time.time()` rather than `or`, which would substitute the current time for a timestamp of `0.0`. Unreachable today, but the `or` was a nearly-faithful port and this one is faithful.

## Three dead locals dropped rather than ported

`error_message`, `timestamp` and `task_duration` were decoded in `_on_workflow_update` and **never used** — dead before this change (F841 sits on the FIB-754 ignore list, so lint does not flag it). Since these lines were being rewritten anyway, they go rather than becoming typed dead code.

Worth a second look from you: the timeline renders all three from the same report and the status bar never has. If it was meant to, that is a feature to add rather than three assignments to keep warm.

## Also verified while re-reviewing

- **Nothing serialises this payload.** Webhooks come from `hooks.py` / `HookContext`, a deliberately separate lifecycle stream (FIB-464), so no `json.dumps` or `asdict` meets the record.
- **`status` is only ever compared with `is AutoLamellaTaskStatus.InProgress`** — the old `"info"` string default was unreachable, and nothing compares it to a string.
- **`update_from_status(None)` now returns cleanly** where it previously raised `AttributeError` inside a Qt slot — that is a qFatal process abort (FIB-329). A small improvement rather than a regression.

**644 passed.**

PR #596, opened 2026-08-27
[refactor] Move the spot burn failure emit into the producer (FIB-824)

patrickcleeve2:
Last of the three. `FibsemSpotBurnWidget.spot_burn_errored` emitted the failure terminal onto the microscope’s signal — but it is wired to **that widget’s own** `FunctionWorker`, so it only ever sees a burn the widget itself started.

## The gap

The unsupervised workflow path never goes through the widget: `tasks/spot_burn.py:183` calls `run_spot_burn` directly. So **a burn that raised there emitted no terminal at all**, and left the status bar mid-run for the rest of the session.

| | before |
| -- | -- |
| base `run_spot_burn` | no `try`/`except` whatsoever |
| Tescan’s | logged and re-raised, without emitting |
| the widget | emitted — but only for burns it started |

Both producers now emit `FAILED` with the exception text before re-raising, and the widget keeps only its UI restore.

**Reporting is not swallowing.** The exception still propagates, and a separate test pins that independently of the report — so a future change cannot quietly turn the emit into a catch.

## The 124-line diff in `microscope.py` is mechanical

The base implementation had no `try` to hang this on, so its body is wrapped. Verified at the AST level rather than by eye:

- **body statements identical once unwrapped** — the `Try` body compared against `main`’s function body, statement for statement
- exactly **one** handler, `except Exception`
- **no `finally`**

## One thing worth your eye — a pre-existing hazard this makes visible

There is deliberately **no `finally`**, which preserves current behaviour: on an exception the restore never runs. But look at what the restore is, and at the comment already on it:

```python
# always restore full frame scanning mode and imaging current
self.set_full_frame_scanning_mode(beam_type=beam_type)
...
self.set_beam_current(current=imaging_current, beam_type=beam_type)
```

It does **not** always restore. A failed burn leaves the beam in **spot scanning mode at the milling current** — and the comment saying otherwise is already wrong on `main`.

I have not changed that here: it is a behaviour change beyond "move the failure emit", and a restore that runs on the failing path can itself raise and mask the original exception, which wants thinking about rather than a quick `finally`. Flagging it rather than fixing it silently, since this PR adds the exact block where such a `finally` would live. Say the word and I will spawn it as its own issue.

**65 passed.** `ruff check` / `ruff format --check` clean.

### DIFF
--- a/fibsem/microscope.py
+++ b/fibsem/microscope.py
@@ -1177,11 +1177,13 @@ class FibsemMicroscope(ABC):
             )
         )
 
-        try:
-            cancelled = False
+        cancelled = False
 
-            # set the beam current to the milling current
-            imaging_current = self.get_beam_current(beam_type=beam_type)
+        # Read before the `try`, so the `finally` below can always restore it. A
+        # failure here means there is nothing to restore anyway.
+        imaging_current = self.get_beam_current(beam_type=beam_type)
+
+        try:
             self.set_beam_current(current=milling_current, beam_type=beam_type)
 
             for i, pt in enumerate(coordinates, 1):
@@ -1231,9 +1233,6 @@ class FibsemMicroscope(ABC):
                     # than one log line later, now that the outcome is recorded.
                     break
 
-            # always restore full frame scanning mode and imaging current
-            self.set_full_frame_scanning_mode(beam_type=beam_type)
-
             # A cancelled burn is not a completed one. Both used to emit `{"finished": True}`,
             # so cancelling rendered "Done" -- the defect the status enum exists to remove.
             self.spot_burn_progress_signal.emit(
@@ -1245,8 +1244,6 @@ class FibsemMicroscope(ABC):
                     total_points=len(coordinates),
                 )
             )
-
-            self.set_beam_current(current=imaging_current, beam_type=beam_type)
         except Exception as e:
             logging.error(f"Error in run_spot_burn: {e}")
             # The failure terminal belongs to the producer. It used to be emitted by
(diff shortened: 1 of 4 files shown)


patrickcleeve2:
Added the `finally` as requested — the restore now runs on the failing path.

Two details worth calling out, because neither is incidental:

**`imaging_current` moved out of the `try`.** A `finally` cannot restore a value assigned inside the block it is cleaning up after — it would `NameError` on exactly the path it exists for. It is now read before the `try`; a failure there means there is nothing to restore anyway.

**Each restore is guarded separately.** An exception raised inside a `finally` **discards the one in flight**, and the original is the one worth having: it says why the burn failed, where a restore failure says only that cleanup also went wrong. Guarding them independently also means neither is skipped because the other failed.

## Tescan needs no equivalent

It burns via DrawBeam with a preset — it sets neither the beam current nor spot scanning mode (it logs and ignores `settings.milling_current`), and its `clear_patterns()` restore was already in a `finally`.

## Both tests checked in both directions

They fail without the change and pass with it. Worth flagging that **the second one was vacuous on first writing**: `set_beam_current` is also called at the start of a burn to select the *milling* current, so a bare `assert_called()` passed whether or not the restore ever ran. It now asserts the *imaging* current specifically, and fails correctly without the fix.

## One ordering change

The terminal report is now emitted before the restores rather than between them. Neither consumer reads the microscope, so nothing observes the difference.

**67 passed**, plus a CI-shaped run with PyQt5 blocked (42 passed, 9 correctly skipped).

PR #647, opened 2026-08-28
[refactor] Send SetImages through a Responder instead of the shared flag

patrickcleeve2:
First conversion of a `workflow_update_signal` blocking site to the typed request/answer primitive that landed in #636 — the `SetImages` instruction, end to end.

## The Qt responder

`QtResponder` (new, `fibsem/applications/autolamella/ui/qt_responder.py`) is the Qt implementation of the workflow's `Responder` protocol. It is **given** the window rather than being it: workflow code is handed a one-method object, and the `QMainWindow` — with every widget and attribute a call site could otherwise reach into — stays on the GUI side of the seam.

`submit(request, future)` is callable from any thread. A queued signal marshals the request to the GUI thread (this object is constructed there, so Qt picks the queued path from the receiver's affinity), a type-keyed handler table dispatches, and the future is completed: `set_result` with the answer, `set_exception` if the widget work failed. That second path is the point — an exception in this widget code used to escape a queued slot, which PyQt5 turns into a process abort with nothing reaching the logfile; now it re-raises on the workflow thread inside `wait_for`, where the task's error handling already exists. An unhandled request type fails the caller immediately with `TypeError` rather than leaving it to the timeout.

One handler so far: `_set_images`, the image block moved verbatim out of `handle_workflow_update`. Handlers grow one per conversion PR.

## The conversion

`set_images_ui` now does:

```python
ask(
    parent_ui.ui_responder,
    SetImages(sem_image=eb_image, fib_image=ib_image),
    abort=lambda: _abort_requested(parent_ui),
    timeout=INSTRUCTION_TIMEOUT_S,
)
```

replacing `WAITING_FOR_UI_UPDATE = True` → emit → `while flag: time.sleep(0.5)`. One future per call, owned by its caller — no other emitter can release it (the shared flag is cleared unconditionally by every emission on the old signal). Instructions are answered by a machine, so silence past 30 s means a wedged GUI thread, not thinking — the timeout lets the workflow unwind and save the record, since Stop is itself a GUI button. `_check_for_abort` is split so the never-raising predicate `_abort_requested` doubles as `ask`'s abort argument: one definition of "stopped" for both mechanisms, and cancellation stays `InterruptedError`, the unwind path call sites already have.

The image branch leaves `handle_workflow_update` in the same PR — `set_images_ui` was its only feeder (verified by grep over the emit sites).

## Tests

`tests/ui/test_qt_responder.py` (new, 6) — all end to end from a real worker thread against the connected Demo harness, the same shape as production:

- images land in the widget, with `WAITING_FOR_UI_UPDATE` untouched throughout
- the widget update runs on the GUI thread
- **a GUI-side failure re-raises on the workflow thread** — previously untestable, because the test process itself would have aborted
- an unhandled request type fails fast, not by timeout
- Stop interrupts a wait the GUI will never answer (the stop event is set only after the wait begins, pinning the wait-loop abort rather than the entry check)
- a wedged GUI thread times out instead of hanging forever

One nonobvious mechanic the wedged tests encode: from the GUI thread itself the responder's connection is direct and dispatch is synchronous, so a wedge is only expressible from a worker thread.

## Notes

- `workflows/ui.py` carried whole-file `ruff format` debt (quotes, spacing); CI lint checks changed files whole, so this PR picks those hunks up.
- Four `F401` dead imports flagged nearby predate this change (their only "uses" on main are words in comments) and are left alone.
- `WAITING_FOR_UI_UPDATE` itself survives until the other two instruction sites (milling config, fluorescence channels) convert.

## Verification

- `pytest tests/ui/test_qt_responder.py tests/ui/test_workflow_update_payload.py tests/autolamella/` — 602 passed.
- `ruff check` and `ruff format --check` clean on the four changed files.

### DIFF
--- a/fibsem/applications/autolamella/ui/AutoLamellaUI.py
+++ b/fibsem/applications/autolamella/ui/AutoLamellaUI.py
@@ -32,6 +32,7 @@ from fibsem import conversions
 from fibsem.applications.autolamella.ui.lamella_name_list_widget import (
     LamellaNameListWidget,
 )
+from fibsem.applications.autolamella.ui.qt_responder import QtResponder
 from fibsem.applications.autolamella.ui.selected_lamella_widget import (
     SelectedLamellaWidget,
 )
@@ -188,6 +189,10 @@ class AutoLamellaUI(QMainWindow):
 
         self._setup_ui()
         self.parent_widget = parent_ui
+
+        # The Qt side of the workflow's Responder seam: workflow code is handed
+        # this one-method object, never the window itself.
+        self.ui_responder = QtResponder(self)
 
         self._protocol_lock = threading.RLock()
 
@@ -1883,24 +1888,8 @@ class AutoLamellaUI(QMainWindow):
                 "No milling task config widget available. Please create a milling task config widget first."
             )
 
-        # update the image viewer
-        sem_image: FibsemImage = info.get("sem_image", None)  # type: ignore
-        if sem_image is not None:
-            self.image_widget.eb_image = sem_image
-            self.image_widget._on_acquire(sem_image)
-            self.image_widget.set_ui_from_settings(
-                image_settings=sem_image.metadata.image_settings,  # type: ignore
-                beam_type=BeamType.ELECTRON,
-            )
-
-        fib_image: FibsemImage = info.get("fib_image", None)  # type: ignore
-        if fib_image is not None:
-            self.image_widget.ib_image = fib_image
-            self.image_widget._on_acquire(fib_image)
-            self.image_widget.set_ui_from_settings(
-                image_settings=fib_image.metadata.image_settings,  # type: ignore
-                beam_type=BeamType.ION,
-            )
+        # Images no longer arrive here: set_images_ui sends a SetImages request
+        # through the Responder seam (QtResponder._set_images).
 
         # what?
         enable_milling = info.get("milling_enabled", None)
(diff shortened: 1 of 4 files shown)


patrickcleeve2:
Two forward-looking findings from review (nothing here fires today):

**1. The Future lifecycle needs a second-writer/late-completion story — and it's cheap to add now.** `wait_for` abandons on abort/timeout without `future.cancel()`, and `QtResponder._dispatch` completes unconditionally. Today that means a timed-out instruction still *executes later* when the GUI unwedges — images written into widgets after the task already failed (no test pins that behaviour either way). And the moment anyone adds the natural hardening — `future.cancel()` on abandon — without guarding `_dispatch`, a crash chain appears: `set_result` on a cancelled future raises `InvalidStateError`, the `except` then calls `set_exception` on the same future, which raises again — and escapes the queued slot into the FIB-329 process abort.

Suggested guard, which pays twice:
- `_dispatch`: `if future.cancelled(): return` before running the handler, and treat `InvalidStateError` from either completion call as "lost the race" (swallow, never re-raise in the slot).
- `wait_for`: call `future.cancel()` when abandoning on abort/timeout, so a stale instruction is discarded rather than applied late.

The same guard is first-writer-wins for the question conversions later — a second completer (Stop button racing an answer, or eventually an agent answering through the Agent Server seam, FIB-851) then loses gracefully by construction. Worth landing before the questions convert rather than after.

**2. `submit` from the GUI thread self-hangs until the timeout.** The queued emit targets this object's own thread; if `ask()` then blocks that thread, `_dispatch` can never run — instructions surface as a misleading 30 s `TimeoutError` ("wedged GUI" when the GUI is just the caller), and the future question types would hang forever. No current caller does it, but ScriptRunner runs data-scripts on the GUI thread, one `set_images_ui` away. One loud guard in `submit` — `if QThread.currentThread() is self.thread(): raise RuntimeError(...)` — turns a mystery hang into a stack trace.

PR #648, opened 2026-08-28
[refactor] Send the milling-config instructions through the Responder

patrickcleeve2:
Stacked on #647. Second and third instruction conversions off the hand-rolled RPC: the two milling-config instructions in `AutoLamellaTask`.

## What moves

- `_set_milling_config_ui` and `clear_milling_config_ui` (`workflows/tasks/base.py`) convert from `WAITING_FOR_UI_UPDATE = True` → emit → sleep-poll to `ask(parent_ui.ui_responder, SetMillingConfig(config) / ClearMillingConfig(), abort=..., timeout=...)` — one future per call, owned by its caller; the instruction timeout and the shared `_abort_requested` predicate from #647.
- Their two branches leave `handle_workflow_update` for `QtResponder` handlers: `_set_milling_config` (load the editor, enable it, front its tab) and `_clear_milling_config`. A missing widget — previously a raise out of a queued slot, i.e. a process abort — now surfaces on the workflow thread as the instruction's failure.

Both dict keys had exactly one emitter each, so the branches die with the conversion.

## Two deliberate calls

- **The transient prompt text is not reproduced.** The old payloads carried `"msg": "Updating Milling Config"` / `"Clearing Milling Config"`, flashed on the instruction label. Matching the SetImages conversion: an instruction carries its content, not display copy — a request must be answerable from the request alone, by any responder.
- **The `deepcopy` at the send is kept.** The editor's copy is the operator's to edit without the task's own config moving under it; the read-back on the question side (`get_config`, a later PR) deepcopies again on the way back.

## Tests

`tests/ui/test_qt_responder.py` +2, same worker-thread-in / GUI-thread-out shape as the rest of the file: a config sent from the workflow thread lands in the editor (`get_config().name` round-trips), the editor's tab is fronted, and the flag is untouched; `ClearMillingConfig` resets the editor to the default config. The wedged/failure/abort behaviours are the seam's, already pinned by the six tests from #647.

## Verification

- `pytest tests/ui/test_qt_responder.py tests/ui/test_workflow_update_payload.py tests/autolamella/` — 604 passed.
- `ruff check` / `ruff format --check` clean on the four changed files; the 32 `F401` dead imports in `base.py` predate this change (same count on main) and are left alone.
- After this, `WAITING_FOR_UI_UPDATE` has two remaining setters: the fluorescence-channels instruction and the embedded clears inside the alignment-area/POI questions.

### DIFF
--- a/fibsem/applications/autolamella/ui/AutoLamellaUI.py
+++ b/fibsem/applications/autolamella/ui/AutoLamellaUI.py
@@ -1939,13 +1939,9 @@ class AutoLamellaUI(QMainWindow):
             self.spot_burn_widget.clear_points_layer()
             self.spot_burn_widget.set_workflow_mode(False)
 
-        milling_config = info.get("milling_config", None)
-        if milling_config is not None:
-            self.milling_task_config_widget.update_from_settings(milling_config)
-            self.milling_task_config_widget.setEnabled(True)
-            self.tabWidget.setCurrentWidget(self.milling_task_config_widget)
-        if info.get("clear_milling_config", False):
-            self.milling_task_config_widget.clear()
+        # Milling config no longer arrives here: the task sends SetMillingConfig /
+        # ClearMillingConfig through the Responder seam (QtResponder).
+
         # fluorescence channel settings
         fluorescence_channel_settings = info.get("fluorescence_channel_settings", None)
         if fluorescence_channel_settings is not None and self.fm_control_widget:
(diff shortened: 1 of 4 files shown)


patrickcleeve2:
One minor from review: **the 30 s instruction timeout can false-fail a task during a genuinely long synchronous GUI-thread stall** (a huge image render on a slow support PC is the realistic case). Modal dialogs are safe — their nested event loops still process queued slots — so this is only the fully-blocked case, and the trade is reasonable; worth one sentence in the `INSTRUCTION_TIMEOUT_S` comment distinguishing the two, since the first person to hit a false timeout will suspect the responder rather than their image size.

Related nit that shows up in this PR: `base.py` importing `INSTRUCTION_TIMEOUT_S` and the private `_abort_requested` from `workflows/ui.py` suggests both really belong in `interaction.py` beside `ask`/`wait_for` — the timeout is a property of the interaction contract, not of the UI helpers, and the underscore import is a smell the move would erase.

PR #659, opened 2026-08-31
[refactor] Give fire-and-forget status its own signal and typed event

patrickcleeve2:
Step 2a of the workflow-signal untangle: the fire-and-forget third of `workflow_update_signal`'s traffic moves to its own channel, typed.

**Why a separate signal:** `handle_workflow_update` reconfigures the interaction UI and unconditionally clears `WAITING_FOR_UI_UPDATE` on every emission — on the shared channel, merely reporting that a task started can release whoever is blocked on that flag. `queue_changed_signal` was split out for exactly this reason; this continues the split.

**What's in it:**
- `WorkflowStatusEvent` — frozen dataclass envelope (`message`, `workflow_info`, `status_bar`, `report`) replacing the bare dict for status traffic.
- `AutoLamellaUI.workflow_status_signal` + `handle_workflow_status` — the display half of what `handle_workflow_update` did for a status payload, and nothing else: no widget-existence guards, no `WAITING_*` flag touches.
- `TaskManager._emit_status` emits the typed event on the new channel.
- Main window: `_on_workflow_update` split into `_on_workflow_status` / `_apply_status_report` / `_refresh_workflow_indicators`; dead `timings`/`t0`/`t1` instrumentation deleted.
- **Latent crash fixed:** `_border_state` was first assigned on the Run click, and the status handler reads it unconditionally — a status payload arriving before any Run was an AttributeError inside a queued slot, i.e. a process abort. Now initialised in `__init__` (the #638 characterisation fixture's workaround for this is removed, so the init is load-bearing under test).

The five status-consuming test files are repointed at the new channel, and the #638 characterisation files gain the typed-path coverage (108 tests across the seven files, plus the 17 responder/confirm tests, all green locally offscreen).

The old dict path stays for everything else; 2b (converting `update_status_ui` and the prompt-clear, then deleting the main window's dict status handling) stacks on this.

### DIFF
--- a/fibsem/applications/autolamella/ui/AutoLamellaMainUI.py
+++ b/fibsem/applications/autolamella/ui/AutoLamellaMainUI.py
@@ -72,6 +72,7 @@ from fibsem.applications.autolamella.ui.workflow_timeline_widget import (
 )
 from fibsem.applications.autolamella.workflows.tasks.queue import QueueOp, QueueResult
 from fibsem.applications.autolamella.workflows.tasks.status import (
+    WorkflowStatusEvent,
     WorkflowStatusUpdate,
 )
 from fibsem.applications.autolamella.workflows.tasks.tasks import get_task_supervision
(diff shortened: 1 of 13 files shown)


patrickcleeve2:
Rebased onto current main (the spot-burn-fix cherry-pick had landed in the same file). Two newer main tests pinned the lamella name-list refresh to `_on_workflow_update` by source inspection; the refresh moved intact into `_apply_status_report` (run for every lifecycle report, same cadence), so those two tests are repointed — plus whole-file ruff format debt they carried, since CI lint checks changed files whole.

PR #664, opened 2026-08-31
[refactor] Run the mill from the Responder, and answer with the config as used

patrickcleeve2:
Fifth question over the Responder seam: `RunMillingTask` — the biggest of the hand-rolled RPCs. Stacks on the PickPOI PR.

**Before:** the mill loop straddled threads four ways — `ask_user(mill=...)` parked on the polled flag; the workflow thread emitted `start_milling_signal` over a `BlockingQueuedConnection`; it sleep-polled `widget.is_milling` across the seam; and it read `get_config()` back at the end.

**After:** one question, and the whole loop lives in `QtResponder` on the GUI thread that owns the widgets:
- show the config in the editor, front the tab, stand the editor's own Run button down (the old `milling_enabled` branch's action, moved into the handler);
- prompt when supervised (`confirm`, from the task's `validate`): **Run Milling** runs the *editor's current* config (operator edits are what actually mill, as before), **Continue** ends the question;
- the widget's own `finished_milling_signal` (success and failure alike, same as the old `is_milling` poll) re-raises the prompt for another round, or completes unsupervised;
- the answer is the config as the editor holds it, with the editor cleared.

Unsupervised runs once unprompted; `enabled=False` only confirms the patterns (single Continue button, no run possible). The workflow thread blocks on one future throughout — no timeout (milling takes as long as it takes), Stop interrupts via the abort predicate at every stage. Headless mode is untouched (runs `run_milling_task` directly).

**Deleted:** `ask_user`'s `mill` variant (the spot-burn variant is now the last one standing) and the `milling_enabled` branch in `handle_workflow_update`.

**Tests** (`tests/ui/test_run_milling_question.py`, 6): `run_milling_task` is stubbed at the milling widget's import site, so runs go through the widget's real thread, button states and finished signal — only the beam time is gone. Run→re-prompt→Continue answers the editor's config with both polled flags untouched; Continue-without-run mills nothing; unsupervised runs once with no prompt; disabled shows confirm-only and the positive click starts no mill; operator edits win; Stop interrupts.

### DIFF
--- a/fibsem/applications/autolamella/ui/AutoLamellaUI.py
+++ b/fibsem/applications/autolamella/ui/AutoLamellaUI.py
@@ -1810,6 +1810,12 @@ class AutoLamellaUI(QMainWindow):
     def _workflow_finished(self):
         """Handle the completion of the workflow."""
         logging.info("Workflow finished.")
+        # Before the early returns: whatever question the run left behind must
+        # come down even if the widgets below are gone. Covers the abort race
+        # where a finished mill re-parks the prompt in the gap before the
+        # aborting waiter cancels its future — by the time this runs, the
+        # workflow thread has exited, so anything still parked belongs to nobody.
+        self.ui_responder.abandon()
         if self.image_widget is None:
             return
         if self.microscope is None:
--- a/fibsem/applications/autolamella/ui/AutoLamellaUI.py
+++ b/fibsem/applications/autolamella/ui/AutoLamellaUI.py
@@ -1885,16 +1885,8 @@ class AutoLamellaUI(QMainWindow):
         # Images no longer arrive here: set_images_ui sends a SetImages request
         # through the Responder seam (QtResponder._set_images).
 
-        # what?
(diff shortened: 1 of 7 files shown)


patrickcleeve2:
Known behaviour change found in bench testing, deliberately not fixed yet: flipping supervision off **while a mill is running** no longer auto-continues when the mill finishes. The old loop re-read `self.validate` on every iteration, so a mid-mill flip meant no re-prompt; the new request snapshots `confirm` at ask time, so `_on_milling_finished` re-parks the prompt from the stale value. Fix direction: carry the supervision state as a predicate (like `abort`) rather than a snapshot, and have the finished handler consult it before re-asking. Applies to the spot-burn question too.

patrickcleeve2:
Same snapshot, other direction (also from bench testing): flipping **auto → supervised** mid-mill used to drop you into the loop — the old code re-read `self.validate` at mill end and prompted. With `confirm=False` snapshotted, an unsupervised run now completes without offering the prompt. The predicate fix covers both directions.

PR #818, opened 2026-09-08
[poses] Derive a lamella's poses through the device axis (FIB-831)

patrickcleeve2:
Closes FIB-831 (PR A of two; PR B re-enables the Default Orientation combo on offset mounts).

## What

A lamella's two poses are now derived with one call each, on either mounting:

```
milling pose      -> fluorescence pose   get_target_position(p, fm.pose_orientation, target_device="FM")
fluorescence pose -> milling pose        get_target_position(p, "MILLING",           target_device="FIBSEM")
```

A compustage takes the device leg with a zero translation and gets the flip it always had. An offset mount gets the traverse, which the orientation-only derivation could never produce, so both of its `poses.py` guards go.

- `_to_milling` requires `get_device_imaging_state("FM", p) is READY` and refuses otherwise, naming why. The `stage_is_compustage` guard is gone.
- `build_lamella_poses` reads the side off the position as "can the objective see the sample from here": the FM orientation on a compustage, as before; parked at the FM on an offset mount. Falls back to the orientation read when there is no FM. A declared `marked_at` is checked against the geometry rather than trusted.
- Asymmetry preserved: `_to_fluorescence` returns `None`, `_to_milling` raises.
- `FluorescenceMicroscope.default_orientation` is initialised from the device declaration (`"FM"` on a compustage, `"FIB"` on the iFLM). New `pose_orientation` falls back to the declared list when the default is not something the objective images from. That is reachable today: loading a saved fluorescence configuration writes its `default_orientation` back onto the combo, on offset too.
- `FluorescenceOverviewGridTask.grid_centre` loses its mounting branch.

## Found while testing the app on the iFLM simulator

Four more commits, each its own fix, all on the offset mount:

- **`move_to_device` arrives on the same piece of sample.** The re-pose leg drove to the orientation by name; `move_to_orientation` rewrites r and t where the stage stands, and the half turn is compucentric about a centre that is not the sample, so the point under the beam was swung away before the traverse (200 µm off on the simulator from an SEM pose). It now drives to the converted position from `get_target_position`, the transform every pose derivation and overview marker already uses. The by-name re-pose survives only as a fallback from a pose the classifier cannot name.
- **Both overview tabs announce a moved pose.** Neither move handler emitted `positions.events.changed`, and writing a pose emits nothing of its own. That signal is the only thing the window re-marks the other overview tab and the lamella cards from, so a move on either canvas left the rest of the window on the old place. Pre-existing on the beam tab too.
- **Holder slots and grid boundaries are drawn where the FM sees them.** Holder places are written in the beams' frame (x = 0, SEM pose); the FM canvas marks fluorescence poses, 48.8 mm away in the FIB pose, so a raw slot landed off the canvas, boundary and all. `stage_context.at_device` carries the place through both legs of the transform, keyed on the device the canvas is for (`"FM"`), not on where the stage is. The beam canvas, no device, and a compustage are untouched.
- **The FM canvas anchors at the FM device's configured origin** rather than wherever the stage was standing when it first drew, which on an offset mount is almost always the beams. Only a place: the pose is still restated from the stage on every draw. A compustage anchors at the chamber origin.

Plus one stale test in `test_simulated_offset_fm.py` that still asserted the FIB-93 refusal (the red CI on the first push).

## What a user on an offset system gets

- A lamella marked from the beam side gets a fluorescence pose (was: silently none).
- *Add Position Here* on the FM overview creates a lamella with a milling pose under the beams (was: refused).
- Dragging a lamella on either overview moves both poses and every view follows.
- Move to FM lands on the marked point.
- The FM overview's slots and boundaries sit under its markers.

## Behaviour changes to know about

- On an offset mount `grid_centre` now snaps the centre to the FM's acquisition pose instead of carrying whatever pose the slot was calibrated in.
- A beam position declared as fluorescence is refused rather than converted. No live caller does this; the FM tab only marks when READY.
- On an offset mount the FM canvas's zero and its holder overlays sit at the FM even while the stage is at the beams. That is the honest picture; before, they were only right by coincidence when the stage happened to be at the beams.

## Not in this PR

Declaring y and z on the device origins. The code reads all three (absent axes read as 0), but the numbers have to be measured: both origins in the same pose (the FIB one, since that is the pose the FM images in), so the compucentric mirror does not end up inside the traverse. A setup flow for that measurement is a separate piece of work.

## Verification

Round trip checked on both simulators. Offscreen: pose, grid FM task, device/imaging-state, move_to_device, connection gate, offset sim, the FM overview widget (267), tab host, tabs-agree, beam-side sync and wiring suites. Ruff check and format clean. Exercised in the running app on the iFLM simulator: add, move, Move to FM, overlays.


patrickcleeve2:
Superseded by the rebuilt FIB-831 stack: #970 (merged), #971 and #972. The full stack is #970 → #971 → #972 → #973 → #974 → #975 → #976 (draft).

PR #835, opened 2026-09-08
[refactor] A configuration loads whether or not it says everything (schema v1)

patrickcleeve2:
Phase 1 of the configuration schema work. It changes no configuration file's meaning
and removes no key — it makes the **readers** tolerant, which is what every later
phase needs and what `rotation_180` needed one-off in FIB-834.

## Two directions, and only one worked

A file carrying *extra* keys has always loaded: `from_dict` reads what it names and
ignores the rest. A file *missing* a key raised `KeyError` — which is why the six dead
`milling:` keys, `imaging.imaging_current` and `stage.manipulator_height_limit` could
not simply be deleted from the shipped files. Every configuration already in the field
would have stopped loading.

Both directions work now, which is what lets the rest of the schema change happen as
one refactor rather than a sequence of them.

- **21 required subscripts become `.get()`** with the field's own default, across
  `StageSystemSettings`, `BeamSystemSettings`, `ManipulatorSystemSettings`,
  `GISSystemSettings`, `SystemSettings` and `MicroscopeSettings`. Every block may now
  be absent, and an empty dictionary loads.
- **`version: 1`** in all eight shipped configurations. Cheap now and impossible to
  add retrospectively: a file without the field is indistinguishable from one written
  before the field existed, so a later format change would have to guess.
- **Unrecognised keys are logged once at load** rather than dropped in silence. Not
  theoretical — `imaging.imaging_current` is in every shipped file, `ImageSettings`
  has no such field, and it has been silently discarded on load for as long as it has
  existed. One line turns "my setting vanished" into "my setting is not supported".
  `sim:` and `protocol:` are deliberately not policed; they carry backend- and
  application-specific keys this schema has no business knowing about.

## Three things the tests found that the audit had not

**A nested reader.** `BeamSettings.from_dict` — reached *through* `BeamSystemSettings`
— also required `voltage`, `hfw`, `resolution` and `dwell_time`. The AST scan that
produced the count of 21 only looked at the classes named in the audit, and a
field-name grep cannot see a class that is handed the whole dict.
`test_an_empty_configuration_loads` is what surfaced it.

**A mutating reader.** `SystemSettings.from_dict` stamped `beam_type` into the
caller's `electron` and `ion` blocks. Harmless while one caller loaded one file; not
harmless once an editor holds a configuration dict and writes it back, because the
stamped key would reach disk. It copies now, pinned by
`test_reading_does_not_mutate_the_dict_it_was_given`.

**A wrong default this change would have hidden** — the second commit, and the one
worth reviewing closely.

## Optional keys turn a loud failure into a quiet one

That is the point of the first commit, but it means a wrong default is no longer
detectable, and there was one. A configuration or a pre-v6 image blob with no `ion:`
block loaded with an ion column tilt of **0 degrees**, because `BeamSystemSettings`
defaulted the field to 0 while `FibsemHardwareGeometry` declared the same physical
constant as 52.

52 degrees is the angle between the columns. Every SEM-to-FIB reprojection turns on
it, so 0 is not a conservative fallback — it is a different instrument, loaded without
complaint. The two declarations could disagree because there were two of them; there
is now one, `DEFAULT_FIB_COLUMN_TILT`, and `BeamSystemSettings.from_dict` picks its
default by beam type rather than using a single number for both columns.

### The test that had to change

`test_system_settings_from_dict_is_not_a_safe_migration_path` asserted the `KeyError`
this refactor removes, so it could not survive as written. Its subject was worth
keeping: it recorded *why* `_geometry_from_legacy_system` reads the raw dict with
`.get()` chains instead of building a `SystemSettings` first — doing the latter would
have inherited the raising on exactly the old files the migration exists to load
(FIB-481).

That reason has genuinely expired rather than been silenced. What replaces it asserts
the property that now matters, and that the shared constant is what makes true: both
read routes recover the same geometry from a blob missing any block.

## Verification

Full non-UI suite: **1 failed, 4921 passed, 22 skipped, 1 xfailed**. The single
failure is `tests/correlation/test_correlation_setup_regressions.py::test_editor_passes_a_real_reference_image`,
which fails identically on stock `origin/main` — checked in a detached worktree, not
inferred.

The new guard was mutation-tested rather than assumed: restoring the flat `0.0`
default fails three tests, and letting the two declarations drift apart fails the
agreement test on the `ion` case alone.

## Not in this PR

Removing keys is phase 2. Two corrections to that plan, found while auditing here:

- The `milling:` block is **not** fully dead — `settings.milling` still has live
  readers in `example/example_milling.py` and `example/slice_and_view.py` (a third is
  in `validation.py`, which #824 deletes). Those examples need updating first.
- `manipulator.rotation` / `.tilt` are live via `is_available("manipulator_rotation")`.
  This PR means phase 2 can drop them from the shipped YAML while the dataclass keeps
  the fields at their defaults; removing the *fields* is phase 3, gated on the
  hardware read.

### DIFF
--- a/fibsem/config/microscope-configuration.yaml
+++ b/fibsem/config/microscope-configuration.yaml
@@ -1,5 +1,6 @@
 # Please don't edit or delet this file, it is the default configuration for the microscope
 # configuration 
+version: 1                                                  # configuration schema version
 info:
     name: openfibsem-microscope-configuration               # a descriptive name for your configuration                     [SPECIFIED]
     ip_address: 192.168.0.1                                 # the ip address of the microscope PC                           [SPECIFIED]
--- a/fibsem/config/odemis-configuration.yaml
+++ b/fibsem/config/odemis-configuration.yaml
@@ -1,5 +1,6 @@
 # Please don't edit or delet this file, it is the default configuration for the microscope
 # configuration 
+version: 1                                                  # configuration schema version
 info:
     name: odemis-microscope-configuration               # a descriptive name for your configuration                     [SPECIFIED]
     ip_address: 192.168.0.1                                 # the ip address of the microscope PC                           [SPECIFIED]
--- a/fibsem/config/sim-arctis-configuration.yaml
+++ b/fibsem/config/sim-arctis-configuration.yaml
@@ -1,3 +1,4 @@
+version: 1                                                  # configuration schema version
 info:
     name: sim-tfs-arctis-configuration               
     ip_address: 192.168.0.1                                 
--- a/fibsem/config/sim-iflm-configuration.yaml
+++ b/fibsem/config/sim-iflm-configuration.yaml
@@ -16,6 +16,7 @@
 # 35 degree pre-tilted shuttle and a half turn to reach the ion beam, which is what
 # puts the FM orientation at (r=180, t=17) -- identical to the FIB one, and the reason
 # an offset system cannot tell the two apart. See FIB-830.
+version: 1                                                  # configuration schema version
 info:
     name: sim-iflm-configuration                            # a descriptive name for your configuration                     [SPECIFIED]
     ip_address: 192.168.0.1                                 # the ip address of the microscope PC                           [SPECIFIED]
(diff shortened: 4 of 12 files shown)


patrickcleeve2:

PR #836, opened 2026-09-08
[refactor] Remove the configuration keys nothing reads (schema v1, phase 2)

patrickcleeve2:
Stacked on #835 — review that first; this diff is against it.

What phase 1 was for. Every key here was verified dead by hand before it went, and an
old file carrying all of them still loads: the readers default, and the keys are named
in one log line rather than dropped in silence.

## Out of all eight shipped configurations

| removed | why it is dead |
|---|---|
| the `milling:` block (6 keys) | read into a `MicroscopeSettings.milling` nothing consulted |
| `imaging.imaging_current` | `ImageSettings` has no such field |
| `stage.manipulator_height_limit` | stored, serialised, never read |
| `manipulator.rotation`, `.tilt` | `false` in every file, which is the default |

**The `milling:` collision is worth a note.** An earlier audit counted 19 reads of
`milling_current` and concluded the block was load-bearing. Those were
`stage.milling.milling_current` — `FibsemMillingStage.milling`, chosen per pattern from
the protocol, a different object that spells its attribute the same way. The
configuration-level one had no readers in `fibsem/` at all. `stage.milling` is
untouched here.

## And one that was never in a file, but was being written into one

`BeamSystemSettings` serves both columns, so `to_dict` put `electron.plasma` and
`electron.plasma_gas` into any configuration saved from the application. An electron
column has no plasma source. They are written for the ion column only now.

`ion.plasma` is untouched and live — `tfs-arctis` sets it with a Xenon gas, and
`is_available("ion_plasma")` reads it. Only the electron side goes.

## The test that had to be rewritten rather than deleted

`test_the_manipulator_keeps_its_own` asserted the manipulator kept the two keys the
stage lost in FIB-834, and said why:

> a line removed from the wrong block would silently give every manipulator an axis it
> does not have, and no other test in the suite would notice

This PR removes those lines from the right block, on purpose, so that assertion
inverts. The hazard it names does not go away though — it moves from the file to the
default, and that is where it is asserted now:
`ManipulatorSystemSettings.from_dict({})` answers `False` to both. `False` is the only
correct default, because an axis assumed present is a move the hardware will refuse.

## Verified, not assumed

The eight files were edited by script and then checked at the **parsed** level against
`git show HEAD:` — every key that differs is one of the intended ones, no value
changed, nothing else moved. Reading the line diff would not have caught a mis-scoped
edit that took `rotation` out of `stage:` instead of `manipulator:`, which is the exact
mistake the test above was written about.

New coverage: no shipped configuration carries a key this version ignores (so another
one cannot accumulate), and a configuration written before the removal still loads with
every removed key reported.

Full non-UI suite: **1 failed, 4927 passed, 22 skipped, 1 xfailed**. The failure is
`test_editor_passes_a_real_reference_image`, which fails identically on stock
`origin/main` and passes on CI.

## Two things reviewers should know

**`fibsem/validation.py:244` still reads `settings.milling`.** Left alone deliberately:
nothing imports that module and #824 deletes it, so editing it here would only make a
conflict. **If #824 is closed rather than merged, that line needs a follow-up.**

**The two `example/` files carry reformat noise.** They had to change — they referenced
the removed field — and `example/` is not in ruff's exclude list while CI
format-checks every changed `.py`, so touching them at all requires formatting the
whole file. The substantive change in each is one or two lines.

## Not in this PR

Writing `stage.devices` and `device_range` into the shipped configurations — the same
phase in the spec, opposite direction. They appear in exactly one file today
(`sim-iflm`), so every other instrument's mounting geometry rides on a dataclass
default nobody can see. It needs no code and is better reviewed on its own.

### DIFF
--- a/example/example_milling.py
+++ b/example/example_milling.py
@@ -22,11 +22,13 @@ The script will:
 
 """
 
+
 def main():
 
     # connect to microscope
-    microscope, settings = utils.setup_session(manufacturer="Demo", ip_address="localhost")
-
+    microscope, settings = utils.setup_session(
+        manufacturer="Demo", ip_address="localhost"
+    )
 
     # setup milling stage (settings and alignment)
     stage = FibsemMillingStage(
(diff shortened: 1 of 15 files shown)


patrickcleeve2:

PR #837, opened 2026-09-08
[refactor] Move SampleHolder into structures.py, unchanged

patrickcleeve2:
Stacked on #836 → #835. Review those first; this diff is against #836.

Groundwork, and nothing else. No behaviour changes, no call site changes, and the four
class bodies are identical before and after.

## Why

`SystemSettings` is about to hold a sample holder, and it cannot from where the class
lives now. `fibsem/microscopes/_stage.py` imports from `structures.py`, so a `holders`
field on `SystemSettings` closes an import loop.

Two ways out: a deferred import inside the function that needs it, or move the class
below the loop. The class moves. A deferred import would leave the cycle in place and
hide it behind an import that has to stay exactly where it is — a thing the next person
has to rediscover, having been given no reason to look.

`SampleGrid`, `SlotCalibration` and `GridSlot` come along. `SampleHolder` is built out
of them; splitting them across two modules to move the minimum would be worse than
moving the group.

`_stage.py` re-exports all four plus `GRID_RADIUS`, so the ~20 modules importing them
from there are untouched:

```
fibsem.structures.SampleHolder is fibsem.microscopes._stage.SampleHolder  →  True
```

## How to review this

**Don't read the diff line by line** — it is a 350-line cut and paste and reading it
proves very little. It was checked by AST instead: the four `ClassDef` nodes were
dumped from `_stage.py` before the move and compared against the ones parsed out of
`structures.py` after.

| class | moved | body identical |
|---|---|---|
| `SampleGrid` | yes | yes |
| `SlotCalibration` | yes | yes |
| `GridSlot` | yes | yes |
| `SampleHolder` | yes | yes |

None still defined in `_stage.py`. A move this size is exactly where a line goes
missing without the diff looking wrong.

Full non-UI suite: **1 failed, 4927 passed** — the same failure on the same test as
before the move, and it fails on stock `origin/main` too.

## What comes next, and one thing found on the way

The substance of the phase — `stage.holders` as a keyed map with `active_holder`, and
the migration that imports an existing `sample-holder.yaml` as one entry — is the next
PR. It is separate because that one carries real risk that this one does not: the
holder file holds **calibrated slot positions**, and a migration that mis-imports
costs a site its calibration rather than failing a test.

Worth flagging for the pre-tilt change after it: `SampleHolder.pre_tilt` **already
exists**, as a property reading back from `self._parent.system.stage.shuttle_pre_tilt`.
Making the holder store `pre_tilt` while `StageSystemSettings.shuttle_pre_tilt` becomes
a property over the active holder means those two point at each other. The existing
property has to go in the same commit as the new field, or it recurses.


patrickcleeve2:

PR #838, opened 2026-09-08
[refactor] The sample holder moves into the microscope configuration (schema v1, phase 4)

patrickcleeve2:
Stacked on #837 → #836 → #835. This diff is against #837.

`stage.holders`, a keyed map, with `active_holder` naming the one on the stage.

The holder was a second file beside the configuration — `sample-holder.yaml`, its own
loader, no version — describing hardware the configuration already describes. A map
rather than a single holder, because a site that swaps a flat shuttle for a pre-tilted
one should select the other entry rather than re-enter its geometry (and, once pre-tilt
moves onto the holder, recalibrate for it).

`loaded_grid` does not come along. Which grid is in which slot changes every session and
keeps its own occupancy file; writing it here would make a saved configuration claim a
grid is loaded that was taken out weeks ago. `to_dict` passes `include_grids=False`.

## The migration is the part worth reviewing

A calibrated holder's slot positions were captured by somebody standing at the
microscope. **They cannot be regenerated, and losing them is worse than anything this
refactor was meant to fix.**

`_resolve_configured_holder` has three cases:

1. the configuration names a holder → use it, touch no files
2. it does not, and `sample-holder.yaml` exists → import it as an entry and select it
3. neither → the shipped default, as before

A calibrated site upgrades with its calibration intact and is never shown a list.

**Nothing is written back.** The imported holder is on `stage.holders`, so the next save
of the configuration carries it — but a session that saves nothing leaves both files
exactly as it found them. Re-importing every session is idempotent and costs nothing;
rewriting a user's configuration on their behalf at connect time is how a good migration
becomes a support ticket.

`StageSystemSettings.from_dict` stays a pure function of its argument and reads no
files. The migration lives at the one place that already touched them.

## A trap found on the way: `apply_configuration`

It replaces `system.stage` wholesale — which is exactly how `rotation` earned its
comment there. The holder is the same trap one field over: a configuration written
before this change has no `holders:`, so pressing Apply would empty the map and clear
the selection, while the running `Stage` carried on using a holder the configuration no
longer knew about.

Which holder is in the shuttle is a physical fact and Apply did not change it, so it is
carried across. A file that *does* name holders wins, because then the user is choosing
one.

## Verification

15 tests. The four failure modes that cost something were **mutation-tested** rather
than assumed:

| mutation | caught by |
|---|---|
| migration silently ignores the site's file | `test_an_existing_holder_file_is_imported_with_its_calibration` |
| imported but not selected | `test_resolving_twice_is_stable` + 2 others |
| migration rewrites the user's file | `test_the_migration_does_not_rewrite_either_file` |
| Apply carry-over removed | `test_applying_a_configuration_keeps_the_holder_selection` |

Full non-UI suite: **1 failed, 4942 passed** — the same pre-existing failure as on
`origin/main`.

## Not in this PR

Pre-tilt moving onto the holder. Note for that one: `SampleHolder.pre_tilt` **already
exists**, as a property reading back from `self._parent.system.stage.shuttle_pre_tilt`.
Giving the holder a stored `pre_tilt` while `StageSystemSettings.shuttle_pre_tilt`
becomes a property over the active holder points the two at each other — the existing
property has to go in the same commit as the new field, or it recurses.

No UI yet for choosing between holders. The map and the selection are readable and
saveable; the picker belongs with the configuration editor.

### DIFF
--- a/fibsem/config/default-sample-holder.yaml
+++ b/fibsem/config/default-sample-holder.yaml
@@ -1,26 +1,16 @@
-# Sample Holder Configuration
-#
-# Slot positions are written by the calibration wizard ("Calibrate slot positions"
-# in the Sample Holder panel on the Microscope tab), which records the orientation
-# and stage geometry they were captured against. A position without that record,
-# or captured against a different pre-tilt / reference rotation, is not trusted
-# and reads as "not calibrated". Do not type positions in by hand.
-
-name: "Pre-Tilted 35deg Shuttle"
-description: "Standard 35 degree pre-tilt shuttle with 2 grid slots"
+name: Pre-Tilted 35deg Shuttle
 capacity: 2
-
 slots:
   Slot-01:
-    name: "Slot-01"
+    name: Slot-01
     index: 0
     position: null
     loaded_grid: null
     calibration: null
-
   Slot-02:
-    name: "Slot-02"
+    name: Slot-02
     index: 1
     position: null
     loaded_grid: null
     calibration: null
+description: Standard 35 degree pre-tilt shuttle with 2 grid slots
--- a/fibsem/microscope.py
+++ b/fibsem/microscope.py
@@ -1109,6 +1109,7 @@ class FibsemMicroscope(ABC):
             self.set_beam_system_settings(system_settings.ion)
 
         if self.is_available("stage"):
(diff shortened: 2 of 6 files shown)


patrickcleeve2:

PR #839, opened 2026-09-08
[refactor] Pre-tilt belongs to the holder, not the stage (schema v1, phase 5)

patrickcleeve2:
Stacked on #838 → #837 → #836 → #835. This diff is against #838.

`stage.shuttle_pre_tilt` becomes a property over the active holder's `pre_tilt`.

Physically correct: swap a 35° shuttle for a flat one and the pre-tilt changes with it.
Today that means editing the stage block by hand and recalibrating — and forgetting the
first silently corrupts every projection made afterwards.

**No call site changes.** `shuttle_pre_tilt` is still a constructor keyword (an
`InitVar` read back through the property), and the setter writes to the active holder,
so the ~25 test files using `microscope.system.stage.shuttle_pre_tilt = 35` as their
setup idiom are untouched. It reads correctly too: the stage's pre-tilt *is* whatever
holder is on it.

## Two ways this could have flattened a 35° site

Both silent, both in the term every projection is built on, and **neither is in the
design this came from** — so they are the part worth reviewing.

**"The holder does not say" is not "flat".** The design returned `0.0` when no holder
was active. But every `StageSystemSettings` built from a configuration is in exactly
that state until `_create_sample_stage` resolves one, and every holder file written
before this is silent on pre-tilt. So `pre_tilt` is `Optional[float]`:

| value | meaning | stage reads |
|---|---|---|
| `None` | the holder does not say | its own configured value |
| `0.0` | a genuinely flat shuttle | `0.0` |

**A holder file may carry a stale `pre_tilt`.** They did once, and it has been *ignored*
ever since the value became derived from the stage — there was a test pinning the
ignoring (`test_from_dict_ignores_old_pre_tilt_key`). Making it authoritative again
would resurrect a number that has not been in effect for however long that file has sat
on disk. A holder loaded from a file takes the configured pre-tilt instead; it becomes
the holder's own from the moment the configuration is saved, which is the point at which
someone has actually seen it.

## The recursion that had to go

`SampleHolder.pre_tilt` was a property reading back from
`_parent.system.stage.shuttle_pre_tilt`. With the stage now reading the holder, the two
point at each other and recurse until the interpreter gives up — **on the first real
connect, not in a test.** The property is replaced by the field in the same commit, and
`test_the_holder_no_longer_reads_the_stage` asserts it stays gone.

`reference_rotation` does not move and stays derived. A reference rotation describes
where the stage's own zero is, and that does not change because a different shuttle was
fitted.

## One disagreement with the design

It expected this to halve `discard_untrusted_positions`, on the grounds that pre-tilt
and slot positions are now co-located and so cannot drift.

They still can. `SlotCalibration.pre_tilt` is a snapshot from calibration time;
`holder.pre_tilt` is current. Editing a holder's pre-tilt *should* discard positions
captured against the old one, which is exactly what the check does. Left intact — happy
to be argued out of this.

## Verification

**Gate: `tests/test_view_corrected_movement.py`** — the forward+inverse sweep.
**360 passed.**

Both hazards above were mutation-tested rather than assumed:

| mutation | result |
|---|---|
| the original design (`return holder.pre_tilt if holder else 0.0`) | 3 tests fail |
| a stale file `pre_tilt` honoured instead of the configured one | 1 test fails |

Full non-UI suite: **1 failed, 4949 passed** — the same pre-existing failure as on
`origin/main`.


patrickcleeve2:

PR #842, opened 2026-09-09
[refactor] What is fitted is asked of the instrument (schema v1, phase 3)

patrickcleeve2:
Stacked on #839 → #838 → #837 → #836 → #835. This diff is against #839.

`manipulator.enabled`, `gis.enabled`, `gis.multichem` and `gis.sputter_coater` were
configuration keys, so a site could describe hardware it does not have, or omit hardware
it does, and nothing would disagree. The same move `rotation` made in FIB-834, one record
over: `_read_hardware_capabilities` fills the fields at connect, and again after
`apply_configuration` replaces the records.

The AutoScript probes are the ones the tree already used at the point of use —
`specimen.manipulator.is_installed`, `gas.list_all_gis_ports()`,
`gas.list_all_multichem_ports()`, and the `hasattr(specimen, "sputter_coater")` check
`run_sputter_coater` makes before it will run. Asked once at connect instead.

## "Cannot say" is a third answer, and it is the important one

Each probe returns `True`, `False`, or `None`. **`None` leaves the field exactly as the
configuration left it.**

The failure here is not symmetric. A subsystem that wrongly *appears* is a menu entry
that errors when used; a subsystem that wrongly *disappears* is a working instrument
that lost a feature on upgrade, silently. So a backend with no way to ask, and a probe
that raises — a missing subsystem and a sick connection raise alike — both mean "leave
it alone", never "not fitted".

## Only the ThermoFisher files lose their keys

| file | capability keys |
|---|---|
| `tfs-aquilos2`, `tfs-arctis`, `tfs-hydra` | removed — probed from the instrument |
| `tescan`, `odemis` | **kept** |
| `microscope` (Demo), `sim-arctis`, `sim-iflm` | **kept** |

`list_all_gis_ports` and `specimen.manipulator.is_installed` are AutoScript APIs.
Nothing here can ask a Tescan or an Odemis what is fitted, so removing their keys would
have disabled the GIS on every one of those systems on the strength of a guess. The
readers still accept the keys everywhere, so a file naming them loads and is not
reported as unrecognised; on AutoScript the instrument's answer simply wins.

## A dispatch key that matched nothing

`is_available` spells it `gis_multichem`. Two call sites asked for `"multichem"` —
[`autoscript.py:3161`](fibsem/microscopes/autoscript.py) deciding which port list to
read, and [`simulator.py:1538`](fibsem/microscopes/simulator.py) deciding whether to use
the multichem. That string matches no branch, so both fell through to `return False`:
the multichem path was unreachable regardless of hardware.

It never surfaced because every shipped configuration set `gis.enabled: true` and the
`gis` branch is tested first at both sites. It would have surfaced the moment the GIS
answer started coming from the instrument, which is this PR. Fixed here, with a
grep-as-a-test so it cannot come back quietly.

## Verification

13 tests. Both guards mutation-tested: reading `None` as "not fitted" fails
`test_a_backend_that_cannot_probe_keeps_the_configured_answer`, and dropping the
`apply_configuration` re-read fails `test_applying_a_configuration_re_reads_the_capabilities`.

Full non-UI suite: **1 failed, 4963 passed** — the same pre-existing failure as on
`origin/main`.

## Not verified on hardware

**The gate for this phase was a bench check and it has not happened.** `is_installed` is
documented on the manipulator, and the GIS calls are the ones already in use, but nothing
here has run against a real AutoScript instrument. The `None`-means-leave-alone rule is
what makes that survivable: the worst case for a wrong probe is a subsystem that appears
when it should not, not one that vanishes.

Worth knowing for later: `is_installed` exists on **13** subsystems, including
`SputterCoater`, `ElectronBeam` and `IonBeam` — more than this PR uses.

### DIFF
--- a/fibsem/config/tfs-aquilos2-configuration.yaml
+++ b/fibsem/config/tfs-aquilos2-configuration.yaml
@@ -33,12 +33,6 @@ ion:
     dwell_time:                 1.0e-06                     # the default ion dwell time
     detector_mode:              SecondaryElectrons          # the detector mode of the ion beam                             [USER]
     detector_type:              ETD                         # the detector type of the ion beam                             [USER]
-manipulator:
-    enabled:                    true                        # manipulator is enabled                                        [USER]
-gis:
-    enabled:                    true                        # gis is enabled                                                [USER]
-    multichem:                  false                        # multichem is enabled                                          [USER]
-    sputter_coater:             false                       # sputter coater is enabled                                     [USER]                        
 imaging:
     beam_type:                  ELECTRON                    # the default imaging beam type (ELECTRON, or ION)              [USER]
     resolution:                 [1536, 1024]                # the default imaging resolution                        [pixel] [USER]
--- a/fibsem/config/tfs-arctis-configuration.yaml
+++ b/fibsem/config/tfs-arctis-configuration.yaml
@@ -33,12 +33,6 @@ ion:
     dwell_time:                 1.0e-06                     # the default ion dwell time
     detector_mode:              SecondaryElectrons          # the detector mode of the ion beam                             [USER]
     detector_type:              ETD                         # the detector type of the ion beam                             [USER]
-manipulator:
-    enabled:                    false                       # manipulator is enabled                                        [USER]
-gis:
-    enabled:                    false                       # gis is enabled                                                [USER]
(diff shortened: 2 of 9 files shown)


patrickcleeve2:

PR #845, opened 2026-09-09
[poses] Two poses, two observations: provenance, no automatic sync

patrickcleeve2:
Stacked on #818. Part of FIB-831 (the issue closes with the last PR of the stack; this one does not reference it as closing).

## The model

A lamella's milling and fluorescence poses are two observations of one piece of sample from two instruments. The transform between them is a first guess for the side nobody has looked through yet, not a coupling, so nothing syncs them behind the user's back any more.

- **Creation asks one question: can the beams mill from this position?** A beam pose at the beams is the milling pose verbatim, whatever tilt it is at, which is what every beam-side caller has always had. Anything else (the FM device on an offset mount, the t = −180 flip on a compustage) gets the milling pose derived first and the fluorescence pose from that. `marked_at` is gone; the FM tab stops overriding the geometry and only decides the objective position to record.
- **The Arctis declares `acquisition_orientations: [FM, SEM]`.** The objective can image at the SEM pose there, which was the original request. Safe under the rule above: a SEM-pose point is the milling pose whichever tab marked it, and its fluorescence pose derives into FM by default or SEM by name.
- **Provenance on the lamella.** `PoseProvenance` observed / derived / stale, serialised as `pose_provenance`; old files load with every pose observed. The plain setters (what workflow tasks use to re-record a pose) record an observation and touch nothing else.
- **No automatic sync.** `sync_fluorescence_pose` is replaced by `derive_fluorescence_pose` / `derive_milling_pose`, explicit overwrites that mark the result derived, and by two Link preferences, `poses.link_fluorescence_position` (beam overview) and `poses.link_milling_position` (FM overview), both on by default. The interactive writers consult them: both overview drags, the lamella list's save-position, and the pose rows in the main window and the coincidence viewer. Off means the other pose is left and marked stale.

## Not in this PR

The checkboxes that expose the Link preferences under the lamella lists, the one-time confirmation before overwriting an observed pose, the provenance icons on lamella rows, the lamella-details derive actions with the orientation picker, and deleting the FM widget's Default Orientation combo. Those follow in the stack.

## Tests

`test_poses.py` rewritten around the creation rule and the derive helpers, with the Arctis SEM cases. New `test_pose_provenance.py`. `test_beam_side_move_syncs_fm_pose.py` now pins the preference both ways. Six device-suite tests used the Arctis at SEM as the "wrong pose" example; they use MILLING now.

Seventeen files: one model change reaches every pose writer, and leaving the sync half-removed would have been worse than the size.


patrickcleeve2:
Superseded by the rebuilt FIB-831 stack: #973. The full stack is #970 → #971 → #972 → #973 → #974 → #975 → #976 (draft).

PR #853, opened 2026-09-09
[overview] A Link checkbox under each lamella list, and one question before overwriting a hand-set pose

patrickcleeve2:
Stacked on #845 (which is stacked on #818). Part of FIB-831; does not close it.

## What

The Link preferences from #845 get their control: a checkbox under the lamella list on each overview tab, named for what a move there will **write**.

- Beam overview: **Link fluorescence position**. Moving a lamella here also derives its fluorescence position from the new milling position.
- FM overview: **Link milling position**. The mirror.

Remembered in preferences. A checkbox rather than a confirmation per drag, which is the kind of dialog people learn to click through without reading.

**One question before overwriting a hand-set pose.** With Link on, the first linked move that would overwrite an *observed* pose asks; a yes holds for the rest of that experiment. Declining does not cancel the move: the dragged pose still moves, the other is left where it was and marked stale. A derived pose is overwritten without asking, since that is a guess replacing a guess. The FM tab already confirms every move, so the warning rides along in that dialog rather than following as a second question.

## Mechanics worth knowing

- The list and the checkbox live in one positions panel, and it is the **panel** the drop path takes back before the overview widget is destroyed. Taking only the list, as before, would leave the checkbox to die with the old widget and hand a dead object to the next build.
- "Once per experiment" is by experiment path for the tab's lifetime. Persisting it in the experiment file was not worth a field.

## Tests

`TestTheLinkCheckbox` in `test_overview_tab_host.py`: labels, preference round trip, panel survives a rebuild, link off marks stale with no dialog, derived overwritten silently, observed asked once, declining keeps the pose and still moves the dragged one, and the FM tab folds the warning into its move dialog. One trap fixed on the way: the stub host recorded every pose as observed, so every linked move opened a real modal dialog, which under the offscreen platform blocks the run indefinitely.


patrickcleeve2:
Superseded by the rebuilt FIB-831 stack: #973 — one rule (a derived pose follows, an observed one stays) replaces the Link checkboxes and the overwrite question. The full stack is #970 → #971 → #972 → #973 → #974 → #975 → #976 (draft).

PR #856, opened 2026-09-09
[lamella] Derive a pose from the other on request, and show where each pose came from

patrickcleeve2:
Stacked on #853 (on #845, on #818). Part of FIB-831; does not close it.

## What

**Derive, on the pose rows.** Each row in the lamella details gets a Derive button that overwrites that pose with one worked out from the other.

- Fluorescence row: into an orientation the FM declares it images from. One orientation derives straight away; more than one (FM or SEM on the Arctis) offers a menu.
- Milling row: into the milling orientation under the beams, then the milling angle is re-read from it.

Every derivation confirms, because every one overwrites, and the question says so when it would overwrite a pose set by hand. The objective position is kept; the dialog says to refocus if the orientation changed. The result is marked derived. Both hosts of the panel, the main window and the coincidence viewer, handle it the same way, and the question they ask is one function in `poses.py`.

**Provenance, visible.** A chip after the position says "derived" or "stale"; an observed pose says nothing, since that is the normal state and a chip on every row is a chip on none. The lamella list gets a matching mark per row: a broken link when a pose is stale, a link when one is derived, nothing when both were centred by hand. That is how the lamellae still to be centred by hand get found.

## Tests

`test_pose_derive_actions.py`: the chips, the single-orientation and menu derive paths, the milling row, the list handing orientations to the fluorescence row only, the list-row mark's precedence, and the main window's handler on the Arctis for fluorescence-into-SEM, milling-from-fluorescence, the observed-overwrite wording, and a refused derivation writing nothing.


patrickcleeve2:
Superseded by the rebuilt FIB-831 stack: #974. The full stack is #970 → #971 → #972 → #973 → #974 → #975 → #976 (draft).

PR #857, opened 2026-09-09
[fm] Drop the Default Orientation control and the instrument-wide default

patrickcleeve2:
Stacked on #856 (on #853, on #845, on #818). Last of the FIB-831 stack.

## What

The FM control widget's **Default Orientation** combo is gone, along with the `default_orientation` attribute on `FluorescenceMicroscope` and the field on `FluorescenceConfiguration`.

It offered a fixed pair, SEM and FM, on every mounting, and a saved fluorescence configuration wrote its own value back, so an offset mount could be handed "FM", a pose its objective never images from. With Derive on the pose rows (#856), an instrument-wide override has no job left: a new lamella's fluorescence pose is derived into the FM's first declared acquisition orientation (`pose_orientation`), and a person chooses another per lamella.

`FluorescenceConfiguration.from_dict` reads past the key an older file may still carry.

## Tests

Seven `test_structures` tests that pinned the field become one that pins the key being ignored; the pose test that pinned the override's fallback now pins the declared default; one FM overview test stops poking the deleted attribute.


patrickcleeve2:
Superseded by the rebuilt FIB-831 stack: #975. The full stack is #970 → #971 → #972 → #973 → #974 → #975 → #976 (draft).

PR #861, opened 2026-09-09
[correlation] Predicted fiducials in the Coordinates tab (FIB-956)

patrickcleeve2:
Stacked on #859 (FIB-881); merge that first. Tracks FIB-956; the review-renderer framing is FIB-957.

## What

The existing correlation widget plus one panel in the **Coordinates** tab, between FM Fiducials and POI: **Predicted Fiducials**.

- **Project FM from FIB** — every FIB fiducial (from the Setup tab's spot-burn seed, or hand-picked) gets a hollow *predicted* FM point through the transform prior. The three that span the pattern best are highlighted amber: drag those first.
- **The prior is a previous run's fitted transform** (`fibsem/correlation/prior.py`): this lamella's latest run, else any other lamella's in the experiment, else the hardware geometry (#859). Rotation and scale only; the translation comes from the placed pairs. The saved METEOR fits agree to under 2° while the geometry sits 5–6° off, which is the difference between 11 px and 72 px after one placed pair (table below).
- Predictions are placed at the slice on screen when you press Project (or the placed pairs' mean slice) and **stay put**; a burn can sit a little off its ring along the depth direction. You scroll to find it and drag.
- A drop confirms the pair (`adjusted` / `user`). Nothing else moves until **Project FM from FIB** is pressed again, which re-places only the untouched predictions through the pairs placed so far: translation from one pair; with the geometry prior the rotation is also refitted from five pairs, with a previous-run prior it is left to the final fit (five coplanar burns cannot pin a rotation better than a full earlier fit). The hint line says what the next press will use.
- **Accept all** turns the remaining predictions into fiducials where they are. They keep `projected` provenance, and the status line says "N pairs placed by you, M accepted from the projection" — accepted-unmoved predictions carry the prior into the fit and are not evidence for it. Only user-placed pairs refine the map.
- Predictions and rejected points **never feed the fit**: `fit_data` (used by the run, the run gate and the stale check) drops them and their index partners; `data` stays complete, so they are saved.
- An FM stack that recorded no geometry (pre-2026-08-04, or odemis-written) can be given an assumed camera transform by the host (`set_assumed_fm_transform`; standalone `--fm-transform none|flip_x|flip_y|flip_xy`), shown in the panel. The library still refuses to guess.
- The standalone launcher offers the lamella's spot-burn pattern on the Setup tab, as the protocol editor does.

The first commit on this branch built the same feature as a separate step-rail widget; live testing showed the rail was the cost and the projection the value, so the second commit reduces it to this. `Coordinate.status` / `.provenance` and the hollow-marker overlay styling from the first commit stay.

The pure logic is `fibsem/correlation/prediction.py` (no Qt), which a `correlation` review renderer (FIB-957) can reuse as is.

## Verified

- `tests/correlation`: 498 passed offscreen (plus the one known local-only failure).
- Driven offscreen on a saved METEOR run with 11 burns and 8 saved picks, prior taken from *another* lamella's run. Residual of the remaining predictions to the saved picks, FM px:

  | placed pairs | geometry prior | previous-run prior |
  |---|---|---|
  | 1 | 72 | 11 |
  | 3 | 52 | 10 |
  | 5 | 40 | 19 |

- `tests/correlation`: 505 passed offscreen (plus the one known local-only failure).

## Known limits

- The first press on a fresh system (no previous run) uses the geometry prior, which on a METEOR is 5–6° off; predictions then plateau at ~50 FM px until five pairs are placed.
- The remaining residual with the previous-run prior is mostly depth: burns on this stack span ~9 slices and one slice is ~4.5 FIB px.
- The protocol editor passes prior runs; an assumed FM camera transform (for odemis stacks) is not yet passed from the microscope config there.
- No inline fit diagnostics, no leave-one-out residuals (FIB-956 follow-ups).


patrickcleeve2:
Added: re-projection never refits the rotation (commit 269bf66f). Measured on the Arctis lamella 02 run: rings 23 px from the burns after 3 pairs, 75 px after 5 with the rotation refitted; same story as the 18° METEOR case. The change is merged forward through #878–#884.

patrickcleeve2:
Added (FIB-979): the first placement learns the previous run's offset, and "Fit Settings" becomes "Method" with one Projection row (rotation source · placement source · Ignore), numbers in the tooltip.

- A completed run records `placement_offset` (fitted minus metadata translation, µm, FIB frame). The next lamella's first projection adds the latest one, skipping runs with rms > 2 µm. A drop still replaces the translation outright.
- Found by the tooltip: the saved Arctis run was an unseeded fit 62° from the geometry (depth column a tenth of the geometry's); as a prior it would have put every ring on the wrong line. A previous run's rotation more than 15° from the geometry is now passed over, and the tooltip says so.
- Merged forward through #878–#884 (three add/add conflicts resolved by keeping both sides).

patrickcleeve2:
Review fix: the placement offset is now measured from placed pairs only (`independent_pairs`). Accepted-unmoved predictions sit at metadata + the previous offset and were echoing it back into the new one. Test updated: a run of accepted pairs records nothing.

PR #867, opened 2026-09-11
[fix] AutoLamellaTaskConfigWidget takes its microscope from the host

patrickcleeve2:
## Summary

Stacked on #865. `AutoLamellaTaskConfigWidget` opened a session of its own at construction, `utils.setup_session()` with no arguments, to give its milling section a microscope. That is a connection to whatever the default configuration names, as a side effect of building a widget, and it was marked with a TODO to take the microscope from the parent.

- The constructor takes `microscope: Optional[FibsemMicroscope]`. The milling section is built only when one is given; the config sync code handles its absence.
- Without a microscope the parameters form builds with no session at all, which is the half the existing tests exercise.
- No tab hosts this widget today: the Protocol and Lamella tabs build the parameters form and the milling viewer separately. The docstring records that, so the next reader knows it is a candidate for removal rather than a live path.

## Verification

- A test monkeypatches `utils.setup_session` to raise and builds the widget: no session is opened, and `milling_task_widget` is `None`.
- A test builds it with a Demo microscope and checks the milling section exists and holds that microscope.
- 71 tests pass across the task-config parameter and form-writeback suites. Lint and format-check clean.

### DIFF
--- a/fibsem/applications/autolamella/ui/autolamella_task_config_widget.py
+++ b/fibsem/applications/autolamella/ui/autolamella_task_config_widget.py
@@ -28,8 +28,8 @@ from PyQt5.QtWidgets import (
 )
 from superqt import QCollapsible
 
-from fibsem import utils
 from fibsem.applications.autolamella.structures import AutoLamellaTaskConfig
+from fibsem.microscope import FibsemMicroscope
 from fibsem.ui.widgets.custom_widgets import TitledPanel, align_form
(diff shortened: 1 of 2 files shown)


patrickcleeve2:
Superseded: the class this fixed had no host, so #869 deletes it instead.

patrickcleeve2:
Correction: the deletion is #870.

PR #877, opened 2026-09-14
[correlation] The fit verdict: leave-one-out, jackknife, rejection (FIB-956)

patrickcleeve2:
Stacked on #861 (which is on #859). Tracks the "Fit verdict" section of FIB-956. Logic only; no UI in this PR.

## What

`fibsem/correlation/verdict.py` (distinct from the existing `fit_diagnostics.py`, the per-point local fitter's diagnostic panel), pure functions over the placed pairs and the prior:

- **`diagnose()`** — per pair: in-fit residual and **leave-one-out error** (refit without the pair, measure where its FM point lands against the pick; the honest error, since the in-fit residual is optimistic). The **target's jackknife** spread (POI through each leave-one-out fit). **Mirror ratio** (RMS seeded from the mirrored prior ÷ chosen fit; near 1 = the pairs cannot say which way is deeper). Depth span, scale vs the pixel-size ratio, the POI's distance outside the fiducials' hull. For a flagged pair, the slice at which it would agree with the others. n + 2 seeded solves, well under a second.
- **`reject_outliers()`** — drop the worst pair while its leave-one-out error is over 2 µm, refit, repeat, keep four. Only the Poor-tier threshold rejects automatically; Check-tier pairs are flagged, never removed, so the verdict never says "have a look" about a pair it already took out.
- **`verdict()`** — good / check / bad, at most two reasons, each naming a fiducial and what to do. No fit vocabulary in the text (a test greps for `mirror`, `leave-one-out`, `jackknife`, `hull`, `pairs`, `RMS`). The Good line reports how well the fiducials agree and does not promise where the target is.

## Calibration

`tests/correlation/fixtures/fit_diagnostics.json`: ten saved runs (pairs, POI, pixel sizes, a prior from another run's isotropic fit; names scrubbed to system + lamella number).

| | pairs | RMS µm | LOO median / max µm | mirror ÷ fit | depth span µm | target jackknife µm |
|---|---|---|---|---|---|---|
| METEOR, 8 runs | 6–13 | 0.17–0.39 | 0.17–0.42 / 0.41–1.46 | 1.45–4.5 | 2.4–4.5 | 0.03–0.45 |
| Arctis run 1 | 7 | 4.99 | 6.6 / 15.5 | **1.10** | 14.9 | 2.7 |
| Arctis run 2 | 7 | 1.91 | 1.9 / 6.0 | **1.21** | 2.9 | 1.2 |

With the thresholds in the module: six METEOR runs **Good**; one **Check** for a target 16 µm outside the fiducials; two **Check** naming one pair each (lamella 13: FM 8 is 1.1 µm off while RMS is 0.30 — RMS alone hides it — and the z-scan puts the burn at slice 36 where the pick is at 32); both Arctis runs **Poor**, ambiguous in depth, worst pairs removed.

Example lines:
- Good fit. The fiducials agree to within 0.2 µm and nothing stands out.
- Check FM 8. FM 8 does not agree with the other fiducials (1.1 µm off). The burn may be at slice 36. Have a look, or remove it.
- Check the target. The target is 16 µm from the nearest fiducial, so the fit is less reliable there. Add a fiducial near the target if you can.
- Poor fit. Do not continue. The fit is ambiguous: the fiducials' depths do not agree with each other, so they cannot say which way is deeper into the sample. Check the z of each fiducial. FM 6 is 15 µm off and was removed.

## Verified

- `tests/correlation/test_fit_verdict.py`: 21 tests — the tier of every saved run, the lamella-13 pair and its suggested slice, the hull case, the Arctis ambiguity and rejection, synthetic damage (one pair moved 4 µm: named and removed), accepted-prediction counting, coplanar wording, the no-fit-vocabulary rule, serialisation.
- `tests/correlation`: 518 passed offscreen (plus the one known local-only failure).

## Not in this PR

Wiring: storing the diagnostics on `CorrelationResult`, the run-bar line, the per-row leave-one-out column, click-to-select. Those follow the coordinate-row redesign in FIB-978 so they are not built twice. The fitter re-fit from a flag (FIB-956) needs the burn-polarity flag first.


patrickcleeve2:
Superseded by #881, which merged this branch and carries the verdict module with the row-index fix from the stack review. Closing so the stack reads as one chain.

PR #878, opened 2026-09-14
[ui] Text roles in tokens; the correlation widgets style from them, not inline (FIB-978)

patrickcleeve2:
Stacked on #861 to avoid conflicts in the tab widget; independent of #877. Section 6 of FIB-978.

## What

The six correlation widget files carried **71 `setStyleSheet` calls**, each choosing its own size and colour: 33 labels at 11 px, 17 at 12 px, one at 10 px, and **34 distinct hex literals** (five greys doing "muted", four ad-hoc tints doing warn/ok), while every file already imported `fibsem.ui.tokens`.

`fibsem/ui/tokens.py` gains text roles as stylesheet fragments: `PANEL_TITLE_STYLE`, `BODY_STYLE`, `BODY_MUTED_STYLE`, `CAPTION_STYLE`, `CAPTION_VALUE_STYLE`, `NUMBER_STYLE` (monospace, so coordinate columns align), `CONTROL_STYLE`, `TABLE_STYLE`, and `state_style(tone, size)` / `state_color(tone)` over ok / warn / error / info / muted.

The widgets name the role instead of the numbers. Greys collapse to `TEXT_MUTED_COLOR` and `BORDER_COLOR`; the RMS badge, RI warnings and fit-dialog tints collapse to the four semantic tokens; the fit-confirmation chip becomes the semantic colour on the panel ground with a matching border. Nothing else changes: no layout, no behaviour.

A guard keeps it that way: `tests/ui/test_correlation_style_tokens.py` greps the six files for hex colours and inline font sizes. The point-type colours in the overlay are the one deliberate exception and are outside the checked set.

## Verified

- Screenshots of the four tabs before and after differ only in the greys and in row numbers being monospace.
- `tests/correlation` + the guard: 508 passed offscreen (plus the one known local-only failure). The colour assertions in `test_correlation_tab_widget.py` now read the tokens.


patrickcleeve2:
Review fix: the tokens pass removed `_WARN_COLOR` from the interpolate dialog but the warn icon still used it, so opening Interpolate with an FM point placed raised NameError (F821 is ignored globally, so CI missed it). Now `WARN_COLOR` from tokens, with a synthetic-stack test that opens the dialog with points placed.

PR #881, opened 2026-09-14
[correlation] The verdict in the run bar, on the rows and on the Results tab (FIB-956, FIB-978)

patrickcleeve2:
Stacked on #880 (rows) and merges #877 (the verdict module). Section 2 of FIB-978, the "Presentation" part of FIB-956's verdict section. Mockup: https://claude.ai/code/artifact/08914ab6-9568-4d46-a4c4-ad7b67b1c154

## What

- **The worker computes the diagnostics** after a seeded fit and stores them on `CorrelationResult.diagnostics` (serialised). An unseeded or undiagnosable fit keeps the old status line.
- **The run bar shows the verdict**: tier-coloured headline, at most two reasons in the sample's terms; a reason naming a fiducial is a link that selects it on both canvases. The RMS badge and the seed note give way to it. A Poor verdict disables Continue.
- **Rows carry their leave-one-out error** after a run (muted; flagged pair amber; over 2 µm red), cleared by the next edit.
- **The Results tab** reports the verdict's evidence in µm (agreement, worst fiducial and its best slice, depth direction, depth span, scale vs pixel sizes, seed distance, pairs) and a per-fiducial table with error and left-out error. Euler angles and the raw translation are gone.

On the METEOR run: *"Check FM 8. FM 8 does not agree with the other fiducials (1.1 µm off). The burn may be at slice 36. Have a look, or remove it."* — the pair the calibration found, with RMS at 0.30 µm saying nothing.

## Verified

Screenshots at 1600×900 (Coordinates after a run, Results). `tests/correlation` + guards: 532 passed offscreen (plus the one known local-only failure); two new widget tests cover the line, the link, the row notes and their clearing, the Results tab, and Continue on a Poor verdict.

## Not here

Automatic rejection is computed but not applied (the line says "Remove it and run again"); the fitter re-fit from a flag and the burn-polarity flag are still FIB-956 follow-ups.


patrickcleeve2:
Review fix: the verdict's pair indices were fit-relative while the run bar mapped them onto the usable FM rows, so after a FIB rejection the leave-one-out note landed on the wrong row and the link selected the wrong point. The worker now hands `diagnose()` each fitted pair's row; `PairDiagnostic.index` and `Reason.pair` are rows, read as rows by the run bar, row notes, link and Results tab. Test: reject FIB 2, then check note/link/Results agree.

PR #885, opened 2026-09-14
[autoscript] Pair the grid on the stage with its home slot by description (AutoScript <= 4.13)

patrickcleeve2:
Seen on the bench today, AutoScript 4.13: with a grid already on the stage, a fresh read shows its home slot as empty in the Sample view. That is what 4.13 reports (the slot reads `Empty`; 4.14 reads `Loaded`, handled in #841), and until now only memory of an exchange we made ourselves kept the grid in its home row.

The slot keeps its `sample_description` while its grid is out, and `autoloader.stage` reports the same name. So an `Empty` slot whose description matches a non-empty stage description is that grid's home: the grid goes back in its home row, present and LOADED, and the working slot holds the same object, so asking to load it is no exchange. No name on either side means nothing is paired and the grid stays "Grid-on-stage", as before.

**Tests**: the pairing on a fresh connect (row present, LOADED, in Slot-04, same object in the working slot, no exchange when asked for); an unnamed stage grid is not paired. 81 across the loader, inventory and Sample view suites.


patrickcleeve2:
Closed after the bench log: on AutoScript 4.13 the description travels with the grid. With the grid on the stage the stage reads Occupied 'name' and its home slot reads a bare Empty, so there is no name on the slot to pair against, and on 4.14 the Loaded state (#841) already answers the question. The gap this tried to close is a 4.13 hardware limit.

PR #893, opened 2026-09-14
[autolamella] A grid's results row puts its overview on the Overview canvas (FIB-71, part 2)

patrickcleeve2:
Second FIB-71 piece, stacked on #892. The hand-off from screening to milling, end to end: from a grid's results to marking lamellae that belong to it.

- **Every results row that recorded an image carries a marker button** ("Mark lamellae on this overview"). It names the image and which canvas it belongs on: the fluorescence one for an FM overview, the beam one for everything else.
- **The Grids tab passes the request up** and the window puts the overview on that canvas, placed by its own recorded position, and brings the Overview tab to the front with a toast saying to right-click it. A fluorescence overview on a system without an FM is refused with the container's own reason.
- With #892, a lamella marked on that overview is given the grid the image says it is of, whether or not that grid is on the stage.

**Tests**: the row's button only where an image was recorded, and what it emits for a beam and an FM overview; in the real window, a request placing the overview on the beam canvas with the grid readable off the record and the Overview tab fronted, and the FM request refused without an FM. 31 across the results widget and Grids tab suites; the overview tab wiring, container and run view suites pass alongside.


patrickcleeve2:
Superseded by the Positions view on the Grids tab: the Overview tab stays the live stage and only shows what is on it.

PR #899, opened 2026-09-14
[correlation] The run bar says what to do next and shows one primary button (FIB-978)

patrickcleeve2:
Stacked on #895. Section 1.4 of FIB-978, and the run bar.

## What

- **The status line names the next step**: load the images → seed the burns → project → drag the rings (n of 4 placed) → place the target → ready to run with n pairs; after an edit, "the points changed since the last run; run again". The verdict line takes over after a run, as before. An action's own line (Project's) still shows until the next edit.
- **One primary button.** Run until there is a live result; then Continue is primary with "Run again" beside it, right-aligned. Continue is not shown before any run.

## Verified

Run bar in seven states on an Arctis lamella (screenshot sheet in the PR comments). `tests/correlation` + guard: 543 passed offscreen (plus the one known local-only failure). One new test walks the seven states.


patrickcleeve2:
Screenshot sheet of the seven states was sent in the session; the states are: no images / images loaded / burns seeded / projected / ready / after run (verdict, Run again + Continue) / edited since the run.

patrickcleeve2:
Added: the z-slider's ‹ › step buttons are 26×24 (were 20×20, reading as 16). Preceded by a format-only commit for fm_canvas.py: the lint job checks formatting of every touched file and this one had never been formatted, so the reformat is separate from the change.

Issue #943, opened 2026-09-17
Coincidence point not being kept for lamellae


equignon:
### Description

Coincidence point used for lamella/overview is incorrect

### Steps

Following the documentation, I've performed these steps:
- Went to SEM Orientation
- Went to coincidence point (e.g. 32.37 mm in my case)
- Acquired an overview
- Added a lamella position

### Additional info

Both the overview and the lamella are using a different Z value (31.88 mm), which makes it impossible to proceed with the protocol as both the SEM and the FIB are looking at very different places. I've might have done something wrong but haven't been able to pinpoint the cause.

fibsem-os version: v0.5.2.dev0

patrickcleeve2:
Hi @equignon 

Thanks for your message.

I've got a few questions to clarify:
1. When you adjusted the coincidence did you do it in xt or FibsemOS? If you moved in xt and then acquired the overview area, the position would be stale, and I think you'd see what you described.
2. Did you add the lamella from the overview image (right click -> add a position) or from the experiment tab -> add a position? Experiment tab uses the current stage position, overview uses the clicked position from the image.
3. When you say the position was looking at very different sem/fib locations, is this at the sem
Orientation or after moving to the milling orientation?


There is an existing bug where the overview area (magenta grid) is stale after changing orientation or coincidence. It will update once you drag it again.

equignon:
Thanks for you answer!

1. I adjusted the coincidence inside FibsemOS and started the overview directly after from FibsemOS too.
2. I added the lamella site by using the overview tab and the right click.
3. After moving to milling orientation. When I clicked on "Move to SEM Orientation", the Z value went back to 32.37 but as soon as it started collecting the overview or processing the lamella site ("Setup Lamella Position" step), the Z value changed to 31.88.

patrickcleeve2:
Thanks, 


I think most likely this is the stale overview grid. If possible could you try restarting while already at sem orientation and coincidence, and reacquire the overview.


From the failed run, could you send the log and experiment file via Help → Report an Issue → private data bundle? If you don't see that menu item, enable it under Edit -> Preferences -> bug report

equignon:
Hi @patrickcleeve2,


Sorry for the late answer, the microscope has been a bit busy these last days.

I've tried to reacquire the overview but even after dragging it, it was still not using the right Z value at the lamella position.
[...]

PR #946, opened 2026-09-18
[overview] A dragged tile grid keeps the gesture, not the position it resolved to (FIB-1007, 2 of 3)

patrickcleeve2:
Second of three on [FIB-1007](https://linear.app/fibsemos/issue/FIB-1007/a-dragged-overview-tile-grid-keeps-its-old-position-z-and-orientation). This is the substantive one. Stacked on #945.

Reported from an instrument: coincidence was set at Z = 32.37 mm, and the overview acquired afterwards — along with every lamella marked on it — came back at 31.88 (#943).

## What was wrong

`_on_grid_moved` resolved the drag through `frame.to_stage`, which is `from_plane` about the **view's origin**, and kept the answer. A canvas is a plane: it can say "500 µm that way along the sample" and it can say nothing at all about height, so the z it returns is the origin's, plus whatever the in-plane offset implies on a pre-tilted stage.

The origin is fixed by the first image placed in the view, or provisionally by wherever the stage was when the tab opened, and is not replaced once real. Change the stage's height after that and a drag made afterwards still plans at the old one — and the runner drives to each tile absolutely, z included, so the run takes the stage back down there. The mosaic records that centre, and a lamella marked by clicking the result inherits it.

This is also why dragging works fine most of the time: with the view anchored at the current height, a drag lands correctly. Re-dragging is therefore *not* a workaround.

## What it does now

`_target` becomes `_drag`, a `DraggedGrid` holding the point in the displayed plane, and `target` resolves it on each read, asking each source what it knows:

- **x and y** from `from_plane` about the origin;
- **z** from `_height_at`, which is the same projection re-based onto the stage;
- **rotation and tilt** from `_posed_like_the_stage`, as before.

**x and y have to come from the origin, not from the stage.** The issue suggests storing the drag as an offset relative to the stage; that breaks the gesture, and no existing test can see it. `_refresh_tile_grid` anchors the overlay at `to_plane(centre, origin)`, so a centre resolved about anything else draws away from the pointer that placed it — measured at 89% of the drag with a provisional origin 12° off the stage's tilt. `test_the_grid_still_lands_under_the_pointer` pins it, and needs an origin at a *different* tilt from the stage or it proves nothing.

Two further changes follow:

- **`acquire` reads the stage once, before the dialog**, rather than planning on the cache. `stage_position_changed` is emitted by `get_stage_position`, so a move nobody polled after never reached the cache (FIB-669) — tolerable for a marker and not for the value a run is driven to. One read on an explicit Acquire press is not the kind of hardware access FIB-544 and FIB-600 are about; `test_opening_the_dialog_costs_no_hardware_read` becomes a count of exactly one, taken before the dialog is built, rather than a ban.
- **A change of acquisition view drops the drag.** The two numbers are expressed in one view's plane, and another beam or orientation is a different plane.

## Verified

- 336 green across `test_overview_widget.py`, `test_overview_confirmation_dialog.py`, `test_overview_tabs_agree.py`, `test_overview_tab_wiring.py`, `test_overview_tab_host.py`, `test_overview_container_tab.py`. All six new tests fail against `main` — checked by reverting and re-running, not assumed.
- **Measured, and it is a trap for anyone extending this:** the drag's dz comes from `shuttle_pre_tilt`, not the stage tilt. On the default configuration (35°): dy −80 µm → dz +56 µm. On the simulated Arctis these tests run against (compustage, `shuttle_pre_tilt: 0`): dz ≡ 0 at SEM, FIB *and* MILLING, despite stage tilts of 0/−128/−23°. So the fixture cannot see the z term at all; both configurations were probed directly.
- Driven in the real application offscreen (`v0.5.2-417-g98e8dc774-dirty`), with the drag going through `TileGridOverlay`'s own matplotlib handlers: anchor (0, 0) → dragged to exactly (220, 140), button enabled, re-centre → target `None`, anchor back at (0, 0).
- The reporter's log and experiment bundle for #943 have not arrived, so the diagnosis of that particular report is still unconfirmed. The defect here is real either way; if that run's `Tiled acquisition centre position:` and `restoring initial position:` both read 32.37, #943 is something else.

Release-Note: An overview acquired after a tile grid was dragged now happens at the stage's current height and orientation, instead of the height the canvas was first anchored at — so setting coincidence and then acquiring no longer takes the stage back to where it was.


patrickcleeve2:
A second field report looks like this bug, and reproduces against `main` exactly:

> When taking an ion tileset image, the stage tilted to the angle that stage was when Autolamella launched. The stage was tilted to an alpha of -17.8 (milling angle 20.2) when Autolamella was launched; the stage was tilted to the desired milling angle (-23 alpha, 15 milling angle); the stage snapped back to -17.8 before the tileset was acquired.

Both tilts are the MILLING orientation, so the displayed view never changes and nothing drops the drag — this is a different path from the re-pose one, and it was not covered. Reproduced on a simulated Arctis at those exact tilts, against `origin/main` and against this branch:

```
### main                                 launch t=-17.80, milling t=-23.00
  undragged, tilt announced              run centre t =  -23.00   OK
  undragged, tilt NOT announced (xT)     run centre t =  -17.80   <-- SNAPPED BACK
  dragged first, tilt announced          run centre t =  -17.80   <-- SNAPPED BACK
  dragged first, tilt NOT announced      run centre t =  -17.80   <-- SNAPPED BACK

### this branch
  all four                               run centre t =  -23.00   OK
```

Two independent causes produce the same symptom, and the report cannot tell them apart because the launch pose is the answer either way:

- the drag froze the pose at the moment it was made, which was the launch tilt;
- a tilt made in the vendor software is never announced, so the cache was still the launch pose (FIB-669).

`test_a_tilt_within_one_orientation_still_reaches_the_run` is added here, parametrised over both, and fails against `main` on both with the reported numbers.

patrickcleeve2:
Superseded — please close rather than merge. Driving #946 in the application found two defects its tests could not see: the first image in a view replaces a provisional anchor and the stored drag offset was then read from the new one (the next run went twice as far), and `_height_at` levelled the stage to the anchor's z, which treats travel along a pre-tilted surface as a height change (z wrong by millimetres with the anchor elsewhere on the slope, and tiles landing beside their own grid). The replacement keeps the drag as a stage position on the map's plane and adds the stage's lift at read time: #963 for main, #961 for the v0.5.3 patch, and #964 is #947 rebuilt on #963. All of #946's behaviour tests are kept there and pass.

PR #947, opened 2026-09-18
[overview] The overview pre-flight says what z will do (FIB-1007, 3 of 3)

patrickcleeve2:
Last of three on [FIB-1007](https://linear.app/fibsemos/issue/FIB-1007/a-dragged-overview-tile-grid-keeps-its-old-position-z-and-orientation). Stacked on #946.

The dialog reported the dragged grid's offset as two components, because two components are what was dragged. The run moves in three: the offset is resolved along the sample surface, and on a pre-tilted stage that surface climbs as the grid travels along it — measured at 0.7 µm of z per micron of y at a 35° pre-tilt, so a half-millimetre drag carries most of that again in z. That is a focus change, and nothing else in the dialog mentions it.

`_target_offset` and `offset` become `(dx, dy, dz)`, and `_centre_text` adds ", and +N µm in z" when there is one to add. Silent below a micron, on the same grounds as the existing x/y dust guard: a clause reading "+0 µm in z" on every drag trains people out of reading the one that matters.

Zero throughout on a compustage, which has no pre-tilt — worth knowing because the simulated Arctis the widget tests run against is one, so the dialog's own tests supply the offsets directly rather than dragging a grid.

The issue also asks for a *warning* when the run centre's z or orientation differs from the live stage. After #946 that cannot happen: both are taken from a live stage read, so there is no second reading to disagree with. A statement of what the stage will do is the useful thing left, and that is what this does.

**On the diff size:** touching either file trips the format-on-touch gate, so most of both diffs is `ruff format` catching up on code this change does not otherwise alter. The substantive part is the `offset` signature, `_centre_text`, `_target_offset`, and three new dialog tests.

**Verified:** 254 green across `test_overview_widget.py` and `test_overview_confirmation_dialog.py`; `ruff check` and `ruff format --check` clean on all four files.

Release-Note: The overview pre-flight dialog now says how far the run will move in z, not only in x and y.


patrickcleeve2:
Superseded — please close rather than merge. Driving #946 in the application found two defects its tests could not see: the first image in a view replaces a provisional anchor and the stored drag offset was then read from the new one (the next run went twice as far), and `_height_at` levelled the stage to the anchor's z, which treats travel along a pre-tilted surface as a height change (z wrong by millimetres with the anchor elsewhere on the slope, and tiles landing beside their own grid). The replacement keeps the drag as a stage position on the map's plane and adds the stage's lift at read time: #963 for main, #961 for the v0.5.3 patch, and #964 is #947 rebuilt on #963. All of #946's behaviour tests are kept there and pass.

PR #960, opened 2026-09-20
[correlation] Optionally re-run the correlation when the points settle (FIB-1020)

patrickcleeve2:
Fixes FIB-1020. Off main.

## Why

Refining fiducials is a loop: drag a pair, read the verdict, drag another. But the feedback you steer by — the verdict sentence, the per-pair left-out error, the reprojected rings — only appears on a press, so the press sits in the middle of the loop. A seeded fit plus its diagnostics is well under a second on seven pairs, cheap enough to be automatic.

## What it looks like

A third checkbox on the Method panel beside **Show diagnostic** and **Auto-accept fits**, off by default and session-only exactly as those two are: **Re-run on change**.

When it is on, an edit schedules a run instead of only marking the result stale:

* **Debounced.** A drag emits `data_changed` on every mouse move, so the run waits `AUTO_RERUN_DELAY_MS` (400 ms) for the points to stop moving, each edit restarting the wait. Measured on the Arctis lamella: twenty edits during a drag produced **0 runs**, and **1 run** once the points settled, with the verdict refreshed to `Check FM 3`.
* **The run gate still decides.** A state the run bar already refuses — no target, mismatched counts, too few pairs — schedules nothing. A timer that fired there would only repeat what the run bar says.
* **A run in flight is waited for, not interrupted.** Killing a `QThread` mid-solve is not worth it for something this short, and a delivered result is checked against the current points regardless (`matches_inputs`, FIB-315).
* Turning the checkbox on with runnable, stale points schedules a run immediately, which is what switching it on plainly means.

**Continue still has to be pressed**, and a poor verdict still refuses it. The auto-run only produces the result; committing the target to the protocol stays deliberate.

## On the placement offset

The issue suggested suppressing `placement_offset` recording for automatic runs, so the "latest run's offset" is not whatever the last twitch produced. Not done, deliberately: `placement_offset_from_runs` already skips any run with rms > 2 µm, so a twitchy intermediate run cannot become the offset a later lamella inherits, and a good-rms automatic run is a perfectly good observation. Adding a second gate would be two rules for one question.

## Verified

Driven on the Arctis lamella 02 run: the drag/settle counts above, and `Continue` enabled after the automatic run. `tests/correlation` + the style guard: 563 passed offscreen. Two new tests cover the debounce and the default-off state, the run gate refusing to schedule, and the in-flight case re-arming instead of starting a second run.


patrickcleeve2:
Made the preference persistent, per your note that off-by-default plus session-only reaches nobody.

`auto_rerun` now lives in `CorrelationConfig` beside `load_spot_burns`, so it is inherited by every lamella in the experiment exactly as the fit methods and RI parameters are. `set_correlation_config` pushes it into the checkbox and `correlation_config` reads the checkbox back. A protocol written before this reads back `False` rather than missing.

Default stays off. Worth flipping once it has had a session on real hardware, which is the point at which we will know whether a verdict per settled drag reads as helpful or as noise.

565 passed offscreen. Two more tests: the config round trip including the older-protocol case, and the preference surviving a trip through the dialog.

